PR #17, opened 2017-10-11
8 styling and theming

### LINKED ISSUE #8: Decide pattern for styling and theming components
Lets figure out how we want to style and provide theming support for components.

We've talked about using `styled-components`, using a `styled` HOC, and/or a `provider` component for theming....among other open source solutions.

We can use this issue to document ideas.

## Acceptance Criteria

- [ ] Initial style implementation in code, with documentation on our plan

Lepozepo:
Fixes #8 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


conrad-vanl:
What would you think if we abstracted out the idea of accessing context from our child components? Im not super fond of random components "knowing" about context. 

For example, perhaps this can be passed in as an argument in the `styled` HOC:

```
StyledButton = styled(({ props, theme }) => ({
   backgroundColor: theme.primaryColor,
})(Button)
```

conrad-vanl:
Is `context` how other provider components work still, I assume? I thought context was still a no-no in react land.

conrad-vanl:
@Lepozepo have you looked at the folder/naming structure on Holtzman? Should we try to replicate that? (I'm looking specifically at calling this folder "Junction"):

https://github.com/NewSpring/Holtzman/tree/master/imports/components

Lepozepo:
@conrad-vanl yeah! Context is used by things like redux, react router, and pretty much anything that has something global that never changes but that wants to be controlled by a component

Lepozepo:
@conrad-vanl I feel like `withTheme` should pass the objects inside `theme` instead of the `theme` object, what do you think?

conrad-vanl:
> @conrad-vanl I feel like withTheme should pass the objects inside theme instead of the theme object, what do you think?

Hmmmm. could there be potential with prop names colliding?

conrad-vanl:
`withTheme` could become `mapThemeToProps` and work just like `mapStateToProps`

Lepozepo:
@conrad-vanl yeah! I think it's cleaner to have `withTheme` just pass the props and I think we could override them (but would discourage it). Passing a mapper function would be pretty good too but I think it should default to passing all available props.

conrad-vanl:
So you didn't like the idea of making `withTheme` become `mapThemeToProps`? I really like the explicitness of it. @burkeshartsis any thoughts on this?

Lepozepo:
I do like the idea, I just think the default should be all props passed down. I'm looking into the implementation right now

Lepozepo:
@conrad-vanl what do you think now?

conrad-vanl:
I like it!

> I just think the default should be all props passed down.

Could you expand a little on why?

Lepozepo:
@conrad-vanl I like having all of it passed down as default so that we don't have to map every time we write a new UI component, I just think it's more clutter. Although I think as we move forward we'll find out whether we want to be inclusive or exclusive.

conrad-vanl:
So my counter to that @Lepozepo is when we're talking about props on a component, I think the verbosity is a good thing. Having to specify which properties of `theme` that a component needs makes it very clear the dependencies of the component, and where that property is coming from.

Since we're now spreading the theme object out - would it be clear to a dev  later on where a random `prop` is coming from:

```jsx
<Button>Hi there!!</Button>

@withTheme
class Button {
  render = () = (
    <Text color={this.props.someColorProp}>My text</Text>
  )
}
// Is it clear where `someColorProp` comes from? 

// What if we have multiple HOCs:
@graphql
@withTheme
@mapStateToProps
class ComplexComponent {
   ...
}
```

I'm fine leaving this like it is for now - just food for thought.

PR #30, opened 2017-10-18
13 environment vars

### LINKED ISSUE #13: Create container for Environment variables
Environment variables can be tricky with create-react-native-app. I've used a `dotenv` type node package before, and it worked...but I've also had problems doing that.

There's an issue on `create-react-native-app` with some more insight:
https://github.com/react-community/create-react-native-app/issues/57

Another way I've solved this for a `crna` project was by using https://direnv.net/ and a `.envrc` file in the project's root. The neat thing about this solution is the `.envrc` is executed, so something like this works:

```bash
export REACT_NATIVE_GRAPHQL_ENDPOINT=http://$(ipconfig getifaddr en0):3010/graphql
```

The con to this is that then anyone would need to install direnv for this to work (or any other way that the developer wants to manage env vars).

## Acceptance criteria

- [ ] Some file in the project (not-tracked by git though) to store env vars
- [ ] Those env vars are loaded into the project and accessible by client-side javascript
- [ ] A plan for handling multiple environments - staging, production, etc

Lepozepo:
Fixes #13 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


Lepozepo:
The current implementation would need something like this to get settings.json working:

```
{
	"REACT_APP_GRAPHQL_URI": "localhost:3001",
	"REACT_NATIVE_GRAPHQL_URI": "localhost:3001"
}
```

Next steps are to have the settings loader automatically differentiate between the platforms and create a helper function to access them so that the file looks flat.

Lepozepo:
All right, this works pretty nice to keep secrets but doesn't have support for multiple environments (we can't really add any environments anyway, both crna and cra hijack `BABEL_ENV` and `NODE_ENV`)

Lepozepo:
## Environment Variables
Create a `settings.json` file with your secrets:
```
{
  "GRAPHQL_URI": "https://secrets.io/graphql"
}
```

Read from settings by importing `@utils/Settings`:
```
import { GRAPHQL_URI } from '@utils/Settings';
```

Lepozepo:
The other path we can take is to just have the settings.json file exist in /src and read it from there.

conrad-vanl:
Okay, now that I understand what's going on here, I like this :)

One question: could `.scripts/run-with-settings.js` use dotenv to parse a `.env` file and export that, so we don't have to use a JSON file?

Lepozepo:
@conrad-vanl I can add support for real environments on a future PR but I think we'll be limited to `development` and `production` only

conrad-vanl:
I'm thinking something like:

```javascript
const config = dotenv.parse(fs.readFileSync('.env'));
```

Lepozepo:
@conrad-vanl Yeah, we can definitely use `dotenv`'s parser and use `.env` files instead. There is a bit of extra flexibility we get from JSON though with nested fields. I'm good with either approach.

conrad-vanl:
With the way your mapping works - would nested fields even work? And I'm leaning .env files just for conformity sake

src/@utils/Settings.js
```diff
@@ -0,0 +1,12 @@
...
+} from 'react-native';
+import pickBy from 'lodash/pickBy';
+import startsWith from 'lodash/startsWith';
+import mapKeys from 'lodash/mapKeys';
+
+const platformPrefix = Platform.OS === 'web' ? 'REACT_APP_' : 'REACT_NATIVE_';
+const configs = pickBy(process.env, (v, k) => (startsWith(k, platformPrefix)));
+const normalizedConfigs = mapKeys(configs, (v, k) => (k.replace(platformPrefix, '')));
+
+export default normalizedConfigs;
```
conrad-vanl: Is there a way where we'd be able to export each of the env vars as named exports, instead of exporting the whole object under default?

```javascript
export { VAR_1, VAR_2 };

# vs 

export default { VAR_1, VAR_2 };
```

(Not sure if possible in this scenario, would just be nice to use named imports)

Lepozepo:
Ah right, it turns it into a string so nested fields would not work. I'll set it up for `.env` with their parser. I think the only point against that could be that JSON is a more widely known format vs `.env`. But `.env` seems to be everywhere anyway, lol.

src/@utils/Settings.js
Lepozepo: I think we can do:
```
module.exports = normalizedConfigs;
export default normalizeConfigs
```

Or something of the sort. What do you want the import statement to look like?

conrad-vanl: `import { MY_CONFIG } from '/@util/Settings'` would be super

conrad-vanl:
```
import CONFIG from '/@util/Settings';

// then later:
CONFIG.MY_CONFIG
```

Is still _okay_ if it ends up being cleaner, FWIW

Lepozepo:
@conrad-vanl this should be gtg. Tried adding support for `.env.development` but I ran into an issue trying to read the environment. There's definitely a little more we can do here but for the sake of progress I think we can merge and I'll open support for multiple environments in a different PR.

PR #34, opened 2017-10-23
use dotenv file for env vars

### LINKED ISSUE #13: Create container for Environment variables

conrad-vanl:
Fixes #13 

@Lepozepo what do you think about using this package - so we can use dotenv files (and rely on less code on our end to support env vars)?

(working on web support)

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,11 +1,11 @@
 {
-  "presets": ["babel-preset-expo"],
+  "presets": ["babel-preset-expo", "react-native-dotenv"],
   "env": {
     "development": {
       "plugins": ["transform-react-jsx-source"]
     },
     "webdevelopment": {
-      "presets": ["react-native"],
+      "presets": ["react-native", "react-native-dotenv"],
       "plugins": [
         "react-native-web/babel",
         "transform-react-jsx-source"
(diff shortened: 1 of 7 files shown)


Lepozepo:
I'm pretty sure `.babelrc` is ignored because we use `react-native-scripts` and `react-scripts`. Did that work for both native and web?

I did have a look at dotenv for sure but while I was testing with it I couldn't get the variables to flow in on both platforms

PR #92, opened 2017-11-09
90.1 apollo direct state manipulation

Lepozepo:
I'm not a fan of this but this is how we could put seemingly global things in apollo's state manager. @conrad-vanl @burkeshartsis what do you guys think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "apollo-cache-inmemory": "^1.1.0",
     "apollo-client": "^2.0.2",
     "apollo-link-http": "^1.1.0",
+    "apollo-link-state": "^0.0.4",
     "expo": "^22.0.0",
     "graphql": "^0.11.7",
     "graphql-tag": "^2.5.0",
--- a/src/@data/Client.js
+++ b/src/@data/Client.js
@@ -1,8 +1,23 @@
 import { ApolloClient } from 'apollo-client';
 import { createHttpLink } from 'apollo-link-http';
 import { InMemoryCache } from 'apollo-cache-inmemory';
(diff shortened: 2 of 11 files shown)


Lepozepo:
Key questions:
- What do we gain?
- What is the cost behind that gain?
- Is it worth it?

mary-pr-poppins:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Warnings</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Please assign someone to merge this PR, and optionally include people who should review.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  No Changelog changes!
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="http://github.com/danger/danger-js/">dangerJS</a>
</p>

conrad-vanl:
Iunno, I kind of like it.....I still think the gain is one single interface for accessing state (graphql/apollo). What do you think the cost is Marcelo?

Lepozepo:
@conrad-vanl here's what I think:

- What do we gain? A global state management tool that uses graphql to handle queries and mutations
- What is the cost behind that gain? An uncommon approach to global state management could make the project less approachable for the open source community but there are few things that will go in there anyway. Also, it creates a bit of an ephemeral dependency for components that will rely on the global state manager but that's also not terrible since that's what happening just by having a global state management solution. 
- Is it worth it? 4 to 5 files for something that can be managed in 1 provider component seems overkill to me but it'll make it easier to bring this data to the server if we ever want to do so

PR #127, opened 2017-11-20
sections menu

### LINKED ISSUE #122: Build Sections menu
## Acceptance Criteria

- [x] It should render full-screen on mobile
- [x] it should function as a dropdown / modal on web
- [x] It should link out to other routes

conrad-vanl:
Fixes #122 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


conrad-vanl:
@Lepozepo I'm going to go ahead and tag this ready for review. I think it might need some more polish still, but would like to get your thoughts on it. 

Essentially - what to test:
- On mobile, tapping "sections" tab item just takes you to the sections list as a tab
- On larger screens, tapping "sections" opens a modal/drop-down menu

Lepozepo:
Ok, I'll have a look in a sec

conrad-vanl:
@Lepozepo yeah, I need to think about this some more, but the behavior you posted in the gif is to be expected and not a bug per say, I think. But there might be a way I can make it not behave that way.

With that said - regarding using a full route change. I'm trying to think through what hits the highest degree of configurability / reusability. A lot of vision on that was cast today that I need to catch you up on...but I think the questions we need to figure out is how much do we want to express the structure of our app in the router?

On mobile - I do think the optimum solution is to have a full page change for this menu. If we build mobile first and build down from mobile - I'm not sure what's the lowest friction way would be to facilitate this functionality in a way that makes sense - in other words - if we express "/sections" as a route on mobile - how do we get away with _not_ expressing it as a route on desktop? We could use `<Link to="/sections" />` on mobile but not on desktop.

Lepozepo:
@conrad-vanl I think it's a design issue we need to bring up with NewSpring. Their current app doesn't support routing to sections but does support changing to landscape (on web) and keeping the visuals consistent. I don't think it's the worst thing in the world to have a button in the bottom navbar that doesn't route anywhere specific, we could keep track of that relatively local state as a query param instead to get the best of both.

Lepozepo:
I accepted it in case you want to merge as is ^_^ it works well anyway

conrad-vanl:
@Lepozepo I pushed a commit that fixes the responsive issue. Let me know what you think.

Still don't want to merge this yet though - there's some tweaks to the API of the AppRouter component I'd like to make.

PR #128, opened 2017-11-27
108 auth headers

### LINKED ISSUE #108: Add Authorization header to graphql requests
## Acceptance Criteria

- [ ] It should add the Authorization header to requests similar to how it is done [here](https://github.com/NewSpring/Holtzman/blob/master/imports/data/graphql/index.js#L40)
- [ ] It uses some form of offline storage to safely store tokens
- [x] Password should be hashed in authorization request [here](https://github.com/meteor/meteor/blob/master/packages/accounts-password/password_client.js#L43)

Created by #106

Lepozepo:
Fixes #108 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


Lepozepo:
I didn't get a chance to build mocks, hope to get to it tomorrow but this should be working now

src/@data/Client.js
```diff
@@ -15,9 +16,11 @@ const ClientStateLink = withClientState({
   },
 });
 
+const httpLink = createHttpLink({ uri: 'https://api.newspring.cc/graphql' });
```
conrad-vanl: Should this URL be loaded from an ENV var?

Lepozepo: Actually yeah! We forgot to do it before (because we didn't have env vars yet). I'll push it with the env and share the env file on slack

Lepozepo: Ah crap, `.env` is server side on web (it's for keys more than config vars). I think I need to build something to get it working. Do we want that? Maybe moving away from .env and using a JSON file with public and private would make things easier.

conrad-vanl: Actually, I think `.env` should be considered public always. On react-native it's all on the client, so technically any keys that are considered "private" would be a no-no on that side.

conrad-vanl:
@Lepozepo should we wait to merge this until after the Heighliner pr is merged, or can I actually test this fully now?

Lepozepo:
@conrad-vanl there's no way to test against their server but I was seeing requests being sent out as expected. I think we should wait.

src/@data/Client.js
Lepozepo: Ok, I'll create an issue to make `.env` public

Lepozepo:
@conrad-vanl this is ready for review, requires 107a on Heighliner to work

conrad-vanl:
Can't seem to get this to work. I pointed the app at the alpha api - `https://alpha-api.newspring.cc/graphql` which I deployed a bit ago (using this branch https://github.com/NewSpring/Heighliner/pull/343).

Actually, looks like an issue on Heighliner:
image

Lepozepo:
@conrad-vanl yeah, I think rock alpha's `sendEmail` doesn't work because the emails aren't configured there (so it fails at that step). I can patch Heighliner to ignore `sendEmail` errors (the emails it does send are weird anyway since they weren't working in the original app)

Lepozepo:
@conrad-vanl are you sure your user isn't already created? I'll add a check for that with a better error

conrad-vanl:
Still having trouble testing this. I try to login - I get no errors - but nothing seems to happen in the UI. Is that expected?

In the heighliner server log - if I use the wrong password -I I see "User not authorized" being logged. So I know I'm using the correct password now. Just nothing happens in app.

Lepozepo:
After login is successful you should see an authToken in local storage and subsequent requests with the authorization header. I didn’t build any side effects on the UI so redirects and those kinds of things aren’t in this PR

conrad-vanl:
Should any of the forms pre-populate with my info when I navigate back to the tab? I thought I saw some `withUser` HOCs being used, which made me think they should.

Lepozepo:
Nope, they’re very bare

Lepozepo:
Awesome, I'll look it over real quick and merge

Lepozepo:
@conrad-vanl looks good. We can split up the forms to different routes in a separate issue.

PR #133, opened 2017-11-29
40 media card component

### LINKED ISSUE #40: Build Media Card Component
**NOTE:** This issue should only encompass the initial component structure. Lets get an MVP version of the card component up with a complete component API - props that support all of the acceptance criteria below. But this issue isn't supposed to make it look pretty or complete.

- [ ] Has a story
- [ ] Has a snapshot test
- [ ] Has a component entry point with props defined
- [ ] Renders something that's recognizable for other development tasks.

Original acceptance criteria and screenshots:

## Acceptance Criteria

- [ ] It should have a heart icon
- [ ] It should have a title
- [ ] It should have a subtitle and icon
- [ ] It should have an image
- [ ] It should support having the image fill the entire media panel
- [ ] It should have a fixed height

## Context
Holtzman/imports/components/@primitives/UI/cards

img_f91801c61601-1
img_99655220222c-1

burkeshartsis:
Fixes #40 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


burkeshartsis:
Looks like neither the ios shadow or android elevation styles are rendered on web. Going to need to make a web specific card wrapper.

src/@primitives/FeedItemCard/index.js
```diff
@@ -64,7 +64,7 @@ const LikeButton = styled(({ theme }) => ({
   paddingBottom: theme.baseUnit,
 }))(TouchableOpacity);
 
-const MediaCard = enhance(({
+const FeedItemCard = enhance(({
```
burkeshartsis: 👍

src/@data/utils/identifyCategory.js
```diff
@@ -30,7 +30,7 @@ export default function identifyCategory(item) {
     default: {
       return {
         ...item,
-        category: category.replace(/(.)(?=.*\1)/g, ''),
+        category,
```
Lepozepo: I have a patch for this [here](https://github.com/NewSpring/Apollos/pull/143)

burkeshartsis:
Ok, I was thinking last night, in order to get the feed cards to be fluid I think we have to do the following,

- either set a static size on the image container and change it with media queries or calculate the images size and set it explicitly. This will allow the card container to adjust to a horizontal layout (image to the right, title to the left) and maintain the images aspect ratio.
- port over [this background image](https://github.com/NewSpring/Holtzman/blob/master/imports/util/backgrounds/background.image.js#L1) util from Holtzman 😢 and add breakpoints
- remove `minHeight` on the card container.
- use media query to set `flexDirection` on card container.

conrad-vanl:
Okay @burkeshartsis played around with this a little locally to better understand issue at hand.

> either set a static size on the image container and change it with media queries or calculate the images size and set it explicitly. This will allow the card container to adjust to a horizontal layout (image to the right, title to the left) and maintain the images aspect ratio.

Lets forget about the horizontal layout on this component for now. They actually never revert to a horizontal card on feed views, only on other screens. Reason why I'm saying this now is the overlay-style card I don't think will work in a horizontal mode. We can build another component later on to handle horizontal layout, or address it when we get to the views that use horizontal layouts.

> port over this background image util from Holtzman 😢

We might have to do this, unfortunately

> either be sure to pass an array of images into CardImage so it can pick the correctly sized image or optionally add breakpoints to the util above so it can pick the image based on the breakpoint (less ideal in my opinion).

I think that util is more concerned about picking an image with the correct aspect ratio. The way I see it, we should favor a square image.

> remove `minHeight` on the card container.

Yes, but this only works if you add an explicit height to `CardImage`, I believe. You could do something like set the width of `CardImage` to '100%', and have an `onLayout` somewhere that watches for layout changes, grabs the card width, passes it to the `CardImage` component and it sets its height to the real width.

> use media query to set flexDirection on card container.

Started playing with this, not that easy, so lets skip ti for now. `FeedView` uses a `FlatList` on web, that does columns by passing it a `numColumns` prop. That should be easy - set `numColumns` using the `mediaQuery` component. On web, we use the `ReactList` component, which from my experimentation so far, doesn't support grid-style views with flex (or columns like `FlatList`). You can set the children to `display: inline-block`, which works. But that's goofy too.

conrad-vanl:
@burkeshartsis pushed up fixes for aspectRatio and such. I think this is likely good to merge -we still have the image issue but I think that is likely a larger issue.

PR #144, opened 2017-12-05
Better theming support

conrad-vanl:
1. Combines @primitives and @modules to a single @ui folder

2. Splits theme up into a few core namespaces:
- colors
- typography
- breakpoints
- spacing
- overrides

Theme ends up looking like:

```js
{
  colors: {
    primary: '#6bac43',
    secondary: '#1c683e',
    tertiary: '#2a4930',
    darkPrimary: '#303030',
    darkSecondary: '#505050',
    darkTertiary: '#858585',
    lightPrimary: '#ffffff',
    lightSecondary: '#f7f7f7',
    lightTertiary: '#dddddd',
    black: '#000',
    white: '#fff',
    transparent: 'transparent',
    error: '#c64f55',
    text: {
      primary: common.darkPrimary,
      secondary: common.darkSecondary,
      tertiary: common.darkTertiary,
    },
    background: {
      default: common.lightPrimary,
      overlay: 'rgba(48,48,48,0.8)',
    },
  },
  typography: {
    baseFontSize: 18,
    baseLineHeight: 20,
    fontFamilySans: null, // force system font
    fontFamilySerif: Platform.OS !== 'android' ? 'Georgia' : 'serif',
    rem: func,
    verticalRyhthm: func,
  },
  breakpoints: {
    xs,
    sm,
    md,
    lg,
  },
  spacing: {
    baseUnit,
    cardBorderRadius,
  },
  helpers: {
    rem: func,
    verticalRyhthm: func,
  },
  overrides: {
    H1: {
      color: 'red'
    }
  },
}
```

`Overrides` allows you to override styles for specific components on a global level. `styled` now supports a second argument that is used as the FQN. Ex:

Say you have an H1 component:

```
const H1 = styled({
  fontColor: 'black',
  fontSize: 32,
}, 'H1')(Text);
```

You can now override the styles for H1 like so:

```
theme = {
  overrides: {
    H1: { fontColor: 'red' },
  },
};

<ThemeProvider theme={createTheme(theme)}>
  <H1>This text is now red!</H1>
</ThemeProvider>
```

Based closely to material-ui-next
https://material-ui-next.com/customization/themes/

Other design considerations:
[NativeBase](https://docs.nativebase.io/Customize.html#Customize)
[Reactackle](http://reactackle-docs.braincrumbs.io/#/theme) (built with styled-components)
[Rebass](http://jxnblk.com/rebass/theming) (also built with styled-components)

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


conrad-vanl:
I'm thinking with these changes `FeedItemCard` would end up looking more like:

```jsx
<FeedItemCard
  backgroundColor={item.colors[0].value}
  textColor={item.isLight ? theme.palette.shades.light.text.default : theme.palette.shades.dark.text.default}
  {...otherProps}
/>
```

src/@ui/CardStack/Transitioner.js
```diff
@@ -41,21 +41,17 @@ class Transitioner extends PureComponent {
     height: PropTypes.number.isRequired,
     direction: PropTypes.oneOf(['horizontal', 'vertical']),
     directionPropNameForChildren: PropTypes.string,
-
-    // from withTheme HOC
-    screenDark: PropTypes.string,
-    screenLight: PropTypes.string,
+    style: PropTypes.any, // eslint-disable-line
```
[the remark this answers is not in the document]
conrad-vanl: PropTypes.object, PropTypes.array, PropTypes.number

Yeah. But then again, this component doesn't really care what `style` looks like,  it just passes it blindly to its child. So, in that sense, I think technically `PropTypes.any` is the most correct, because as far as this component is concerned, it don't care none.

conrad-vanl:
Alrighty @burkeshartsis, this should be officially g2g now

src/@ui/typography/H2/__snapshots__/H2.tests.js.snap
```diff
@@ -8,7 +8,7 @@ exports[`the H2 component should accept additional props 1`] = `
   ellipsizeMode="tail"
   style={
     Object {
-      "color": "#505050",
+      "color": "#303030",
```
burkeshartsis: Looks like you might have changed the default font color. The default font color is actually `theme.colors.secondary`. You can either change the type component references to use `.colors.secondary` or change swap the primary and secondary text colors in the type file (the latter is probably more appropriate).

src/@ui/theme/defaultTheme.js
```diff
@@ -0,0 +1,143 @@
+import { Platform } from 'react-native';
+import * as baseTypes from './types';
```
burkeshartsis: I think `themeTypes` might be more helpfully descriptive

src/@ui/theme/defaultTheme.js
```diff
@@ -0,0 +1,143 @@
...
+// The available theme types get included into the theme,
+// and the active theme type gets merged into the theme.
+// This allows for switching of theme values between different color palettes.
+// The theme types included by default are "light" and "dark".
+// But, you could add your own theme types (ex: "kids", with a more vibrant set of colors).
+// Most of the colors that a component uses should come from the active type, not from the
+// base colors above. For example, notice below how `shadows` gets the shadowColor from
+// `colors.shadows`, which is provided by the active theme type.
+export const type = 'light';
+export const types = baseTypes;
```
burkeshartsis: I get that you export them here so they are in context with the rest of the theme but why do these have to be reassigned?

src/@ui/theme/defaultTheme.js
```diff
@@ -0,0 +1,143 @@
...
+
+// The available theme types get included into the theme,
+// and the active theme type gets merged into the theme.
+// This allows for switching of theme values between different color palettes.
+// The theme types included by default are "light" and "dark".
+// But, you could add your own theme types (ex: "kids", with a more vibrant set of colors).
+// Most of the colors that a component uses should come from the active type, not from the
+// base colors above. For example, notice below how `shadows` gets the shadowColor from
+// `colors.shadows`, which is provided by the active theme type.
+export const type = 'light';
```
burkeshartsis: I think `defaultType` might be more descriptively helpful.

src/@ui/typography/H2/__snapshots__/H2.tests.js.snap
conrad-vanl: @burkeshartsis if default set to secondary, all of the text in the app appear much lighter then what's in their current app, do they not? Like looking at card headings and such. Maybe body copy and smaller UIText are using the lighter color

burkeshartsis: So I think you're correct. Card titles are darker than content titles. We might just need to go over all the existing type component styles/usage and make sure we are being faithful/accurate. But in looking at the default junction styles they are lighter than what you have here.

conrad-vanl:
> Left some comments and found a few regressions. I like that we merged the `@primitives` and `@modules` here but I feel like we might have gone to far with the "everything under one roof" idea. `/@ui` feels like a presentational folder. It's components. It's stuff you see. I feel like we might still have a case for a `/@primitives` (or similar) folder to keep HOC's that `/@ui` uses (theme, styled, router, fontLoader, etc.). With that in mind, we might also want to think about moving the `defaultTheme` and types out of `/theme` where all the work is happening. Maybe rename `/theme` to `createTheme` and keep a `/theme` in `/@ui` or even at root that devs would interact with. Thoughts?

I don't know. I like the simplicity of just throwing everything ui related in a single folder so you don't have to think about where it goes. I think your example still introduce some subtle complexities...like right now `theme` exports two components: `<ThemeProvider />` and `<ThemeMixin />` (as well as a `withThemeMixin` HOC). Under your example, would we pull those out and put them in `@ui` and `@primitives` separately? If `@primitives` are for HOCs, should everything in `@data` go there too? 

While I agree with what you're saying in principle, I kinda feel like keeping everything presentational related under one subfolder is the simplest thing to do for now, and if that breaks down with scale we can re-organize it once we have a clear picture on how our code is categorized (with scale).

src/@ui/theme/defaultTheme.js
conrad-vanl: disagree, i think. everything in this folder are the "defaults" and get composed as the "default theme" - https://github.com/NewSpring/Apollos/pull/144/files#diff-1b4f2c38552e852db30f5b0808ba41f9R4

If I changed this to `defaultType` I would have to selectively handle it in this file, and change it back to just `type`

src/@ui/theme/defaultTheme.js
conrad-vanl: fixed.

conrad-vanl:
Aight @burkeshartsis , I think I responded to all your comments or fixed them. Let me know your thoughts. I think for typography we can follow up and fix the color values in a separate PR, this atleast puts it to a cleaner state (`text.primary` maps to `colors.lightPrimary` or `colors.darkPrimary`)

PR #147, opened 2017-12-11
Content view

### LINKED ISSUE #148: Hook up Content Views
## Acceptance Criteria

- [x] It should link content cards to content views
- [x] Get most data on screen (in unstyled way)
- [x] Open issues for additional components that need built (such as video, audio, tab view in devotionals, etc)

## Context

Given [CONTEXT] when [ACTION] then [RESULT]

conrad-vanl:
Fixes #148 

Adds content views for most types of contents, will be opening up issues for additional components that need built.

This is not supposed to be a finalized content view PR, just a "lets get content views in the app, get a framework for flushing out content views together, and then move on" PR.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [x] Review new test logic


conrad-vanl:
Requesting reviews from either @Lepozepo or @burkeshartsis .... whoever wants to take this (or both!).

src/@ui/HTMLView/styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from '@ui/styled';
+import { Text } from 'react-native';
+
+const Link = styled(({ theme }) => ({
+  color: theme.colors.text.link,
```
burkeshartsis: How does this look on the web? Do we need to worry/think about `:visited`

burkeshartsis: I tested this and it doesn't seem to be a problem. A `:hover` underline for web would be a nice touch though.

src/@ui/Header/index.js
```diff
@@ -16,28 +17,36 @@ const StyledHeaderBar = styled({
...
 }), 'Header.Text')(H6);
 
 const enhance = compose(
+  withTheme(),
   pure,
   branch(() => Platform.OS === 'web', renderNothing),
   setPropTypes({
     titleText: PropTypes.string,
     backButton: PropTypes.bool,
+    barStyle: PropTypes.oneOf(['light-content', 'dark-content']),
```
burkeshartsis: Would it be better to standardize around the usage of `isLight`? Seems like this is the same thing.

src/@ui/Paragraph.js
```diff
@@ -0,0 +1,8 @@
+import { View } from 'react-native';
+import styled from '@ui/styled';
+
+const Paragraph = styled(({ theme }) => ({
+  paddingVertical: theme.sizing.baseUnit / 2,
+}))(View);
+
+export default Paragraph;
```
burkeshartsis: This feels like a typography component. Should this live in `@ui/typography/`?

src/@ui/ContentView/ByLine.js
```diff
@@ -0,0 +1,20 @@
...
+const enhance = compose(
+  pure,
+  branch(({ authors }) => !authors || !authors.length, renderNothing),
+);
+
+const ByLine = enhance(({ authors }) => (
+  <H6>By: {startCase(authors.join(', '))}</H6>
+));
+
+ByLine.propTypes = {
```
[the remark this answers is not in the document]
conrad-vanl: Hmm...I thought we landed on using react-native standards for setting `propTypes` and `defaultProps` ? Or am I remembering incorrectly 🤔 

Doing `Component.propTypes =` like this, and the `enhance` function above, makes sure that these statics are set on the prototype of the object being exported, which is what I think we really want. Using the recompose HOCs for propTypes and defaultProps works, but it ends up gobbling up clarity on what object/component they're actually being defined.

src/@ui/ContentView/ByLine.js
```diff
@@ -0,0 +1,20 @@
...
+  authors: PropTypes.arrayOf(PropTypes.string),
+};
+
+export default ByLine;
```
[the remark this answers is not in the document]
conrad-vanl: Seemed to have too much logic in it to keep it inside of `ContentView`. If we're looking at stories as documentation, stories for components that are internal to a component (not meant to be used by other components) are overkill IMO... need to document the components we want other people to use, not components we don't want them to use. 

Test coverage should be picked up by tests  on `ContentView` itself (which I don't think there is currently...I'll work on that), but I'm not opposed to adding test coverage for this component as well.

src/@ui/HTMLView/styles.js
conrad-vanl: Agreed. I think that can be a part of #149

src/@ui/Paragraph.js
conrad-vanl: I went back and forth. I figured it'd get moved in #149 which is why I just put it in the root level (no folder, no tests). It's a block level element (uses `View`), so putting it inside of `typography` might be inaccurate IMO.

Potentially renaming it to `ParagraphView` could help. Again, something that I think belongs as part of #149

src/@ui/Header/index.js
conrad-vanl: Good question. I think so, maybe? Let me explain original thought process:

I chose to standardize this around the same prop in React Native's `StatusBar`. The reason is when I first implemented this, I actually got the theme values mixed up and was rendering the status bar dark when it should be light, and visa versa. It turned out it was because on NewSpring, the header is by default dark even though the base theme is light. However, on branded content, we want the header to match the theme of the content (when isLight==true, the header isLight as well). The rest of the time in the app, the header doesn't follow that logic.

So I separated the naming of the header a bit and based it around what I felt was a more precise description (at the time) - which was naming the style after the same prop used in `StatusBar`, and then getting header to match the style of the status bar.

However, at this point, I don't really see the reason/argument anymore, and don't mind changing it to `isLight`. I agree that could be more consistent.

conrad-vanl:
> It also looks like the single views (a single story, devotion, news, etc.) for a lot of these sections are basically identical. I wouldn't mind exploring how we could simplify this further.

What were you thinking here?

WET vs DRY.

We talked about this a few times actually, but potentially at a slightly higher level. I think alot of these views are going to have more differences as we flush them out.... such as Tabs on Devotions, videos on some, list of child/sibling content items on some, similar content lists on others, etc. If we abstracted these views much more (put more into the `ContentView`), I'm afraid we'd end up with a lot of conditionals and crap in `ContentView` - which is what I believe you said before you didn't want to happen.

I think the potential to dry these views up will likely be revealed once we flush more of this out. For now, the repetition i think keeps the flexibility consistent, and is closer to how Holtzman is currently, actually.

src/@ui/ContentView/ByLine.js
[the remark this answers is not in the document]
burkeshartsis: I'm not debating the merits of one approach or the other just that I wanted to be consistent. Last I head was we wanted to keep everything inside `compose` so that we could express the entity of a components API in one place. Happy to go either way here.

src/@ui/Header/index.js
burkeshartsis: FWIW, I think `isLight` is terribly named but it's the lemons we were given.

src/@ui/Paragraph.js
burkeshartsis: Happy to address this in #149

conrad-vanl:
@burkeshartsis added tests!

src/@ui/ConnectedImage/index.js
```diff
@@ -48,37 +73,38 @@ class ConnectedImage extends PureComponent {
...
-      if (sizeCache[source.size]) return Promise.resolve(source);
-      return (new Promise((resolve, reject) => {
-        Image.getSize(source.uri, (width, height) => resolve({
-          width, height,
-        }), reject);
-      })).then((sizeForCache) => {
-        sizeCache[source.size] = sizeForCache;
-      });
-    }));
+  get aspectRatio() {
```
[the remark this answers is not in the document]
burkeshartsis: Updated tests in 51ad36e to address this.

PR #152, opened 2017-12-11
Feed card fixes

burkeshartsis:
Patches a bunch of stuff that got missed in #133.

- fixes card sizing by removing flex all over and letting the images aspect ratio define card size.
- fixes image sometimes not filling parent due to minHeight on card and flex all over.
- **re-adds missing `isLiked` prop after error during previous merge conflict.**
- fixes stories not using images prop.
- adding missing states to some tests and stories.
- updates CardWrapper to match changes made to feed card down stream and swaps it into feed card as it's card component.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/src/@ui/CardWrapper/index.js
+++ b/src/@ui/CardWrapper/index.js
@@ -17,23 +17,7 @@ const StyledCard = styled(({ theme, cardColor }) => ({
   width: '100%',
   backgroundColor: cardColor || theme.colors.lightPrimary,
   borderRadius: theme.sizing.borderRadius,
-  ...Platform.select({
-    ios: {
-      shadowColor: theme.colors.lightTertiary,
-      shadowOffset: {
-        width: 0,
-        height: 1,
-      },
-      shadowOpacity: 1,
-      shadowRadius: 3,
-    },
-    android: {
-      elevation: 3,
-    },
-    web: {
-      boxShadow: `0 1px 4px ${theme.colors.lightTertiary}`,
-    },
-  }),
+  ...Platform.select(theme.shadows.default),
 }))(View);
 
 /*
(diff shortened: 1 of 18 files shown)


src/@ui/CardWrapper/index.js
```diff
@@ -7,17 +7,19 @@ import styled from '@ui/styled';
...
-const StyledCard = styled(({ theme }) => ({
-  backgroundColor: theme.colors.background.default,
+const StyledCard = styled(({ theme, cardColor }) => ({
+  width: '100%',
+  backgroundColor: cardColor || theme.colors.lightPrimary,
   borderRadius: theme.sizing.borderRadius,
   ...Platform.select({
     ios: {
-      shadowColor: theme.colors.shadows.default,
+      shadowColor: theme.colors.lightTertiary,
```
conrad-vanl: Can we reference a shadow color in `theme.colors.shadows` instead (perhaps add the needed color to the `shadows` object)

src/@ui/CardWrapper/index.js
```diff
@@ -26,23 +28,31 @@ const StyledCard = styled(({ theme }) => ({
       shadowRadius: 3,
     },
     android: {
-      elevation: 1,
+      elevation: 3,
+    },
+    web: {
+      boxShadow: `0 1px 4px ${theme.colors.lightTertiary}`,
```
conrad-vanl: Ditto. see above

src/@ui/CardWrapper/index.js
burkeshartsis: great catch. Not sure why these were not updated with the last theme PR. No worries. Updated to use `...Platform.select(theme.shadows.default)` which I find pretty trick 💥

PR #182, opened 2017-12-12
form inputs, buttons, and chips! 🐿

### LINKED ISSUE #167: Create+style form inputs
## Acceptance Criteria

- [ ] It should create form inputs (or reutilize ones we've already built)
- [ ] It should style form inputs
- [ ] It should migrate existing views to use these form inputs
- [ ] It should add documentation of these inputs to storybook

## Components from Holtzman:

- [ ] Multi Select (KeywordSelect)
- [x] Single Select
- [x] Checkbox
- [ ] Date
- [ ] Fieldset
- [ ] File
- [ ] Form
- [x] Input
- [x] Label
- [x] Select
- [ ] TagSelect
- [x] TextArea

conrad-vanl:
Fixes #167

Adds 3 input primitives: `Input`, `Picker`, `Switch` that can be used to create most types of text entry inputs (email, password, text, etc), dropdown menus (with `Picker`), and toggles/checkboxes (with `Switch`).

Also adds a base `Touchable` component that utilizes the best feedback mechanism for the OS (essentially uses native feedback on Android), and both a `Button` component and a `Chip` component ([a-la material design](https://material.io/guidelines/components/chips.html)).

The `Chip` component will be used in place of what they call "Tags" in their current codebase, it should be able to be used to create the search functionality in the group finder rather easily.

One thing you might consider missing from this PR is error handling and validation. I'm thinking we might consider using this library, and eventually adding an error state to inputs: https://github.com/jaredpalmer/formik

But here are some other options (that may or may not be compatible with react-native-web: https://github.com/jondot/awesome-react-native#forms)

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


src/@ui/inputs/withInputControlStyles.js
```diff
@@ -0,0 +1,8 @@
+import styled from '@ui/styled';
+
+const withInputControlStyles = styled(({ theme }) => ({
+  paddingVertical: theme.sizing.baseUnit / 4,
+  height: 30 + (theme.sizing.baseUnit / 2),
```
Lepozepo: I think this is missing `outline: none` on web to remove the default `:focus` outline

conrad-vanl:
> I think it would reduce a lot of setState boilerplate

Didn't catch that it was uncontroller. Kinda sucks, but that's okay for now I think. As far as boilerplate goes - I was thinking we could maybe use something like Formik? https://github.com/jaredpalmer/formik

src/@ui/inputs/withInputControlStyles.js
conrad-vanl: fixed

conrad-vanl:
@Lepozepo fixed the merge conflict and removed the outline on web.

Let me know your thoughts on using formik for forms.

Lepozepo:
Formik looks pretty cool ^_^ I'm ok with using something like that too

PR #188, opened 2017-12-19
Loading states for feed items.

### LINKED ISSUE #115: Loading/error states for Media Card
## Acceptance Criteria

- [ ] MediaCard component should have a loading state (see app)
- [ ] Should also have some sort of error state (?)

burkeshartsis:
Fixes #115. `ConnectedImage` has a placeholder but it doesn't appear to be ever rendering it in app (works fine with fake data in storybook). I'm guessing that in the real world it's too busy validating the status of each image which is blocking rendering. Probably just need to move some code out of render.

TODO: 
- [x] Need to pass `FeedView`'s `isLoading` to the `defaultFeedItemRenderer` to trigger loading states. 
- [x] Fix warnings from placeholder returning multiple children.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


src/@ui/ConnectedImage/index.js
```diff
@@ -95,16 +102,22 @@ class ConnectedImage extends PureComponent {
 
   render() {
     let { source } = this.state;
-    if (!Array.isArray(source)) source = [source]; // TODO: could render a loading image instead?
-    if (!every(this.state.source, image => image.width && image.height)) return null;
+    if (!Array.isArray(source)) source = [source];
+    if (every(source, image => image.width && image.height)) {
+      this.isLoading = false;
```
conrad-vanl: Two questions:
1. Why is this a instance property on the class? Why not just `const isLoading = !every(source, image => image.width && image.height)`. Doesn't appear to be accessed outside of the render function.
2. If we want to keep it an instance property, would it be cleaner to make it a getter function and get rid of the changes to the constructor above?

Like:

```
get isLoading() {
  return !every(this.state.source, image => image.width && image.height);
}
```

burkeshartsis: I think originally I was planning on creating a class method or I was going to add an `isLoading` prop and set `this.isLoading` to equal that so that an image's loading state was tied to both it's source request and the loading status of it's parent (like a card) so the whole card could fade in together. We decided in favor of progressive loading and this was just a carry over. I do think there is space to refactor `render()` to move some of the logic out.

Updated to use your suggestion.

PR #208, opened 2018-01-03
154 content children

### LINKED ISSUE #154: Content children
Create a horizontal card list component, add it to content views with children to render a list of content children.

## Acceptance Criteria

- [ ] It should list content children on content pages with children (series/single, study entry/single, etc)

## Context

Given [CONTEXT] when [ACTION] then [RESULT]

burkeshartsis:
Fixes #154 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


conrad-vanl:
@burkeshartsis regarding the first point in my review above, I think I was incorrect. It looks like their query is just different. It appears when getting children, they are filtering the children but the channel type. So on series they are only showing sermons:

https://github.com/NewSpring/Holtzman/blob/master/imports/pages/series/VideoList.js#L71

PR #210, opened 2018-01-05
Preserve state on transitions

conrad-vanl:
The primary result of this PR is that scroll-position and component state is preserved when you navigate back to a view. For example - you scroll down on the home feed, tap a card, then go back - the feed will be right off where you left it.

This is achieved by keeping components mounted that you navigate away from within a singular card stack. For example - say you start on the home feed and then open a devotional card. Both the devotion and home feed are still mounted. You go back to the home feed, and the devotional card is unmounted. This might sound expensive - but in practice this is how react-navigation (and navigation-experimental) both work. The other benefit is that this way _all_ component state is maintained...you start watching a video, pause it, tap something else, go back - your video is still paused. 

There's extra safe-guards in place (for web mainly, but works on mobile too) that keeps the currently mounted stack size limited to just one stack of cards. For example - say you navigate into a series, then a sermon. The stack size would be 3 ('/', 'series/:id', 'series/:id/sermon'). On web, you could tap on the "home" button from this screen which would usually just push '/' onto the stack...but because of how `<CardStack>` works, it knows this is a new "stack" of cards, and resets the mounted component stack.

Another benefit to this structure is that back transitions, especially the drag to go back gesture, are more performant.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


Lepozepo:
Awesome!

PR #227, opened 2018-01-18
Content header gradients

burkeshartsis:
Adds header gradients to Series and Devotional headers.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


src/@ui/ContentView/__snapshots__/ContentView.tests.js.snap
```diff
@@ -121,3 +121,311 @@ exports[`the ContentView component renders 1`] = `
...
+            this is some cool html
+          </Text>
+        </Text>
+      </View>
+    </View>
+  </View>
+</View>
+`;
+
+exports[`the ContentView component renders with video 1`] = `
```
burkeshartsis: This test doesn't appear to be rendering correctly? Thoughts?

conrad-vanl: Hmm...maybe the `WebView` component is stubbed out in Jest?

burkeshartsis: I went ahead and merged this and created an issue #233 to track this concern.

PR #242, opened 2018-01-24
Fix logic error with using `&&` in react-tree

### LINKED ISSUE #239: Article html parse error
## Acceptance Criteria

- [ ] HTML content should render error free.
- [ ] HTML parser gracefully handle invalid HTML.

## Context

Looks like the new "Why Diversity Should Matter To You" article is throwing a parsing error. 

simulator screen shot - iphone x - 2018-01-22 at 15 18 29

conrad-vanl:
Fixes #239

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/src/@ui/ContentView/ContentMedia.js
+++ b/src/@ui/ContentView/ContentMedia.js
@@ -69,14 +69,14 @@ const ContentMedia = ({
     visual = (
       <View>
         <ImageHeader source={images} />
-        {imageOverlayColor && (
+        {imageOverlayColor ? (
           <ImageOverlay
             colors={getGradientValues(imageOverlayColor).colors}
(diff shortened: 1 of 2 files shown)


burkeshartsis:
Also let's not start branch names with `&&-` though. Us command line folks cringe a bit 😬

burkeshartsis:
@conrad-vanl It looks like there are a few other files that also use the `&& ( //...` conditional. I don't believe they are subject to the same issue as they are not checking a string though. Can you take a look at `[src/series/Single.js](https://github.com/NewSpring/Apollos/blob/%26%26-bug/src/series/Single.js#L73)` which checks both an object and a string. If you want this changed let's do it now. Otherwise feel free to merge.

PR #243, opened 2018-01-24
237 give subnav

### LINKED ISSUE #237: Build give subnav
- [ ] it should use `TabView`
- [ ] it should navigate between `Dashboard`, `Now`, and `History`

Lepozepo:
Fixes #237 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


Lepozepo:
newspring

Lepozepo:
newspring

src/@ui/forms/EditSavedPaymentMethodForm.js
```diff
@@ -63,6 +63,7 @@ export class EditSavedPaymentMethodForm extends PureComponent {
   };
 
   render() {
+    console.log(this.props.isValid);
```
conrad-vanl: can probably remove this 😉

src/give/GiveRoutes/TabViewPagerPanPatched.js
```diff
@@ -0,0 +1,278 @@
+/* @flow */
```
conrad-vanl: Can you talk me through the purpose of this and the patch?

src/give/GiveRoutes/index.js
```diff
@@ -0,0 +1,104 @@
...
+    return (
+      <TabView
+        initialIndex={this.currentRouteIndex}
+        routes={this.props.routes}
+        renderScene={this.props.scenes}
+        renderHeader={this.renderHeader}
+        onChangeFinished={Platform.OS === 'web' && this.handleOnChangeTab}
+        // NOTE: We need to use TabViewPagerPan to be able to return to the
+        // correct index position
+        renderPager={props => (<TabViewPagerPan {...props} />)}
```
conrad-vanl: If we're having to do a bunch of weird stuff to support using `TabView` with our router, I almost wonder if we should just not use it, and just let these pages be plain routes without the swipe animation and what not.

src/give/GiveRoutes/TabViewPagerPanPatched.js
Lepozepo: This is just so that `onChangeFinished` runs after the animation is complete. Their current handler has inconsistent behavior when you tap a Tab link vs when you swipe to it, this just makes it consistent (it is however not entirely necessary now that I'm running `replaceState` to keep track of history without triggering rerenders)

src/give/GiveRoutes/index.js
Lepozepo: From what I noticed during exploratory testing and from reading through their source it looks like `navigationState.index` isn't always respected on their ios/android implementation, specifically when you route into give, then out, and then back in, even if the index is set to anything other than 0 the pager will resolve `position` to 0 or -0. `TabViewPagerPan` is always consistent though and with the addition of `useNativeDriver`, the performance is decent I think.

src/give/GiveRoutes/TabViewPagerPanPatched.js
conrad-vanl: If not entirely necessary, do you think we could remove this patch then?

Lepozepo:
@conrad-vanl this is ready for re-review unless you'd like me to remove `TabViewPagerPanPatched` and stick with their current `onChange` implementation or to bring in their current `TabViewPager`'s for iOS and Android and patch their initial index issues

conrad-vanl:
@Lepozepo I'd rather not have `TabViewPagerPanPatched` in our source. It doesn't match the rest of the project's coding standards and is a big blob of code smell. If we can refactor so that it's not needed, lets do that.

Lepozepo:
@conrad-vanl ok, the impact it'll have on routing is minor anyway, I'll patch in a sec

Lepozepo:
@conrad-vanl I've removed it, this should be good to go now :D

conrad-vanl:
This is looking pretty good. However, giving is starting to diverge quite a bit style-wise on web. What's your plan here?

screen shot 2018-01-25 at 5 44 36 pm

Lepozepo:
@conrad-vanl I haven't gotten to desktop views yet but the plan is to use mediaQueries and reuse the chart component in a separate issue

Lepozepo:
I have a patch coming in to make links smooth like I think you want them :D

src/give/Dashboard.js
```diff
@@ -5,7 +5,7 @@ import {
   TouchableWithoutFeedback,
 } from 'react-native';
 import PropTypes from 'prop-types';
-import { compose, branch, renderComponent } from 'recompose';
+import { compose, branch, renderComponent, getContext, withProps } from 'recompose';
```
conrad-vanl: This is great! Instead of using `getContext` directly though, could `TabView` export a `withTabViewActions` HOC or something, that injects `jumpTo` as a prop? So component authors don't have to interface with context directly?

conrad-vanl:
@Lepozepo approved. Feel free to merge! :)

PR #253, opened 2018-01-26
Custom Typography Styles

### LINKED ISSUE #149: Style basic typography in content
`HTMLView` (`@ui/HTMLView`) and `ContentHTMLView` (`@ui/ContentHTMLView`) drive the html-to-react conversion for showing any body text within the different content types. This issue aims to create the appropriate components needed (such as `Li`, `BlockQuote` or `Paragraph`) and styling them so that these views match the existing design language.

## Acceptance Criteria

- [ ] Content bodies within the different content sections match existing designs
- [ ] Title, subtitle, author bylines match existing designs

burkeshartsis:
Initial work for #149 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [x] Review new test logic


src/@ui/FontLoader/Asset.js
```diff
@@ -0,0 +1,185 @@
+/* eslint-disable */
```
conrad-vanl: Lets add a note about why this file is here and where it came from. Maybe change the URL to be very clear ("PatchedAssetFromExpo.js" or the like)

burkeshartsis: K. I changed the file name but not the exports that way when Expo gets their act together all we have to do is delete this file and change the import path in `Font.js`

package.json
```diff
@@ -23,7 +23,7 @@
     "eslint-plugin-jest": "21.2.0",
     "eslint-plugin-jsx-a11y": "6.0.2",
     "eslint-plugin-react": "7.4.0",
-    "exp": "^47.4.0",
+    "exp": "^49.0.3",
```
burkeshartsis: Weird... how did this not get updated with everything else?

conrad-vanl: Not sure. It's not needed in this PR, but I was running out and wanted to test on my phone, was hopeful this would get the expo CI build to pass :-D

src/@ui/FontLoader/Font.js
```diff
@@ -0,0 +1,126 @@
+// @flow
```
conrad-vanl: Can we add a comment on where this file is from and why it's here?

src/@ui/FontLoader/PatchedAssetFromExpo.js
```diff
@@ -0,0 +1,185 @@
+/* eslint-disable */
```
conrad-vanl: Ditto. Can we add a comment on where this file is from and why its here?

conrad-vanl:
Text inputs are bold in the current app, but pretty light here:

img_0570

These two headings got squashed:

img_0572

How should we do links/toucahbles so we don't end up with mixed type faces?

img_0571

The top of "Psalms" is cut off:

img_0575

PR #254, opened 2018-01-26
upgrade deps

conrad-vanl:
Fixes [issue-number]

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


src/@ui/styled/createStyleSheet.js
```diff
@@ -36,7 +36,7 @@ const createStyleSheet = (stylesToGenerate) => {
     styles[index] = generatedStyle;
   });
 
-  if (styles.length === 1) [styles] = styles;
+  if (styles.length === 1) styles = styles[0]; // eslint-disable-line
```
conrad-vanl: FWIW, this was causing some sort of build error. I think we do array destructuring/assignment like this elsewhere and it seemed to work fine, but this one was failing for some reason. Maybe related to metro boilder.

PR #263, opened 2018-01-30
Production settings

IsaacHardy:
- Updates `app.json` with newspring specific application settings for deploying to the App Store and Google Play Store. 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


app.json
```diff
@@ -1,8 +1,21 @@
 {
   "expo": {
+    "name": "NewSpring Apollos",
```
conrad-vanl: I think this might should just be "NewSpring" - this will end up being the app name on a user's home screen

IsaacHardy: That sounds good!

IsaacHardy:
@conrad-vanl @burkeshartsis @Lepozepo I've added most of our build config to app.json. I've launched an alpha version on the Google Play store and everything looks great! 

Please review the changes I've made and let me know if there is anything I need to update/change.

IsaacHardy:
@conrad-vanl I would be quite open to bring these build versions down. Currently on android, you cannot create a new release of a build version that is lower than the version of the release that is above. (i.e. alpha must have a higher version than beta; beta must be higher than production;)

On iOS, I don't believe there are any version requirements.

conrad-vanl:
@IsaacHardy yeah, I was able to confirm that. For some reason I thought we were able to get android version numbers to reset for some reason. Oh well.

on iOS - I believe you're correct.

Feel free to merge as is! Or, we can change the `buildNumber` on iOS - I believe it can just be "0" and if we need to make multiple builds for a given release, we can increment it then. `buildNumber` is tied to the version, unlike Android's `versionCode`.

IsaacHardy:
@conrad-vanl I'll update that `buildNumber` on iOS, first.

PR #265, opened 2018-01-30
web feature disablement

### LINKED ISSUE #215: Disable features on web to match Holtzman
Holtzman on the web today has a different home page, and many of the section links link out to the newspring website. We should duplicate this feature,

conrad-vanl:
Fixes #215 by adding in redirects to live website

Also adds a `Hero` component that can be used for composing these split-panel views:

https://cl.ly/3z0y1R21281u

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


src/AppRouter.js
```diff
@@ -26,6 +26,8 @@ import Settings, { ProfileDetails, ProfileAddress, ChangePassword } from './sett
 
 import { Results as GroupFinderResults, GroupSingle } from './group-finder';
 
+const redirectToNewspring = path => window.location.replace(`https://newspring.cc/${path}`);
```
Lepozepo: Should this be `window.location.href = ...` instead? So that when they hit back they go back to the app instead of one page before that.

conrad-vanl: @Lepozepo I don't believe so. Because this is fired inside of a `<Route>`. The way it happens is the app links you to say `/locations`, and `/locations` then redirects you to `https://newspring.cc/locations`. 

This is supposed to happen similar to a classic `301` redirect. Which is what this does effectively.

conrad-vanl: If I used `window.location.href` and you hit the back button you'd end up on a blank page. Or possibly a redirect loop.

Lepozepo: Ah, that makes sense

conrad-vanl:
Lets merge the typography PR first, since its larger

Issue #250, opened 2018-01-25
Missing Nav Bar on section lists (articles, sermons, devotionals, etc...)


IsaacHardy:
## Acceptance Criteria

- [ ] Nav bar should be visible on all section list views (articles, sermons, devotionals, etc...)

conrad-vanl:
This issue is more complex due to the universal router. I have a plan to simplify how routing works which would make this much easier to implement, if it can afford to wait. Last time I tried to tackle this issue I ran into a lot of bugs/complexities with the current card stack transitioner implementation. Which led me to read into more how iOS does card stack navigators internally and I realized I was over complicating our current implementation

PR #266, opened 2018-01-31
style TabBar to better match current app

conrad-vanl:
This branch was originally going to contain a fix for #250, but #250 seems to be a tough nut to crack with our current router / CardStack setup. I'll update that issue with more details.

Instead, this styles the TabBar on web to more closely match the current app:

This PR:

image

my.newspring.cc:

image

What's on master:

image

Type changes in #253 makes this even better :)

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


src/@ui/TabBar/index.js
```diff
@@ -22,9 +22,10 @@ const styles = StyleSheet.create({
 });
 
 const TabBar = compose(
-  styled(({ theme }) => ({
-    backgroundColor: theme.colors.darkPrimary,
-  }), 'TabBar'),
+  styled({
+    // todo - this color was hardcoded in Holtzman and has no corresponding theme value
+    backgroundColor: '#202020',
```
conrad-vanl: Came from: https://github.com/NewSpring/Holtzman/blob/f4559b70df0a17a753fa2415a4966ff95f91a102/imports/data/store/nav/reducer.js#L129

conrad-vanl:
@burkeshartsis I'm going to think about this some more and potentially end up patching #250 while i'm at it.

If you have thoughts on the current changes though-  namely what's in 5527af6 - i'd love to hear it.

conrad-vanl:
@burkeshartsis I'm going to throw this back in to for review. I'll create a separate branch for the other nav stuff I want to do.

If you want me to throw this theme value - `#202020` into the theme - let me know what you think it should be named. I'm at a blank.

src/@ui/TabBar/index.js
burkeshartsis: Could we use `Color.darken` instead? This would allow us to keep things tied to the current Theme and avoid hardcoding a color.

src/@ui/TabBar/Link.js
```diff
@@ -19,17 +19,18 @@ const styles = StyleSheet.create({
     height: 50,
   },
   largeButton: {
-    height: 80,
+    height: 75,
+    transform: [{ scale: 0.9 }],
```
burkeshartsis: This is not ideal and we both know why. For now, can you add a comment explaining this with a TODO to address it later with a designer.

burkeshartsis: Also if this were pure css I would say use `scale3d` so that it's drawn on a gpu layer so that every time there is a need to rescale it's accelerated. Is there any way to accommodate that here?

burkeshartsis: fixed in debefdc

IsaacHardy:
@conrad-vanl What needs to be updated with #250?

PR #273, opened 2018-02-01
189 native end in web

### LINKED ISSUE #189: on iOS: Donation must finish inside web browser
## Acceptance Criteria

- [ ] 21.2  The collection of charitable donations must be done via a web site in Safari or an SMS
- [ ] Current app lets you get to the last step, then essentially does something like "your contribution must be finished in a web browser", you tap a button, it opens safari with a pre-authenticated URL and your contribution is on screen. You press finish and you're done.

Lepozepo:
Fixes #189 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


Lepozepo:
This isn't ready yet, deep links had a few poorly documented caveats but should be good tomorrow now that they're working consistently and both users and contributions are resumable

Lepozepo:
Oh, one more thing I need to add is an ENV for controlling the initial redirect, but here's the preview:

newspring

Lepozepo:
@conrad-vanl this is ready to review

conrad-vanl:
@Lepozepo looking good!!

One question: I notice this is using an in-app browser (which I guess you got to work - I couldn't get it to work before). However, I think we need to kick the user out to the actual browser, not the in-app browser.

IsaacHardy:
@conrad-vanl Is the need to use the actual browser instead of `in-app-browser` because of Apple's guidelines or a design suggestion?

conrad-vanl:
@IsaacHardy my understanding of apple guidelines:

> 21) Charities and contributions Apps that include the ability to make donations to recognized charitable organizations must be free The collection of donations must be done via a web site in Safari or an SMS

conrad-vanl:
@IsaacHardy since apple's policy isn't very clear, we can always try it like it is, worse case they deny and we have to change it.

src/@ui/NativeWebRouter/DeepLinking.js
```diff
@@ -0,0 +1,51 @@
...
+  static contextTypes = {
+    router: PropTypes.shape({
+      history: PropTypes.shape({
+        push: PropTypes.func.isRequired,
+      }).isRequired,
+    }).isRequired,
+  };
+
+  async componentDidMount() {
+    this.push('/give/now');
```
conrad-vanl: Why does this push to `/give/now` on mount?

Lepozepo:
@conrad-vanl @IsaacHardy I replicated the flow for what we had in Holtzman (I may be missing some copy here and there) but it was definitely using an in app browser [here](https://github.com/NewSpring/Holtzman/blob/master/imports/util/inAppLink.js) with a fallback to safari (I think we get the fallback for free on expo but I need to double check).

src/@ui/NativeWebRouter/DeepLinking.js
Lepozepo: Oh, that wasn't supposed to be there, lol

PR #274, opened 2018-02-02
149 UIText Deprecation

burkeshartsis:
Partial work towards #149 

Deprecates and removes `UIText`. Most references were replaced with `H7` or `H6` but should not be considered accurate replacements and will need further review. This also refactors `ButtonLink` and removes duplicate components that served the same purpose.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


IsaacHardy:
@burkeshartsis Just out of curiosity, why was there the need to deprecate UIText component?

burkeshartsis:
@IsaacHardy #FunFact It was actually never part of the NewSpring design guidelines. Originally we created to so we would have a default sans serif (read: Colfax) text component as a compliment to `BodyText` (serif, Droid Sans). We also crutched on it when we didn't really know what typography component we should use in. Now that typography is sorted out and we are pulling the app into shape this component shouldn't really exist any more.

IsaacHardy:
Ah! That makes sense. Thanks for the details!

PR #277, opened 2018-02-02
Consistent tab views

conrad-vanl:
Makes all the tab views in app consistent.

@Lepozepo I also changed the Pager on android for giving to `TabViewPagerAndroid`. Performance still isn't great - it looks like there is some blocker in the giving dashboard that is killing performance. Even just scrolling on that screen is bad. Don't worry about tracking it down yet - with Android it's usually something dumb, like having an obscenely large border radius or creative use of a prop like https://facebook.github.io/react-native/docs/view.html#rendertohardwaretextureandroid

I just need to know if `TabViewPagerAndroid` will work okay. It appears that everything still works as far as navigation, but perhaps i'm missing something.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [x] Review new test logic


Lepozepo:
We should be good with `TabViewPagerAndroid` but I feel like the root of the issue could be in the way currentRoute works. I think it maybe causing remounts or rerenders as we're tracking it and passing it down via `initialIndex`

conrad-vanl:
Yeah, could be. I'll take a look at this when I get into more giving stuff. does the rest of this PR look good?

conrad-vanl:
@burkeshartsis or @Lepozepo whoever can get to this first to review, thanks!

conrad-vanl:
I'm waiting for my android to build up enough charge to test, but I believe 198be05 should fix this error you're experiencing.

burkeshartsis:
Looks like there are different styles for the tab view being used in different places. I'm assuming this is linked to theme types as that's how the tab styles appear to be working in devotionals (see images below)
### devotional tab styling (Android)
image
image
However, on Give the tab bar appears to be using a darkened primary green or worse, that weird blue overlay color that android tries to put on stuff. 
### give tab styling
image
I like the idea of bringing the tab bars into theme styling so it seems like Give should match the devotionals.

conrad-vanl:
@burkeshartsis what you are seeing was intended. I guess what you're saying is it looks bad.

The default styles is to use `colors.primary` for the background color, and `colors.secondary` for the indicator. The default on web is to use white background and `colors.primary` for the indicator. This is so that the tabview matches the style of the header wherever it's used.

On devotionals it uses the devotional's primary color, and then calls darken or lighten on it for the indicator.

PR #281, opened 2018-02-02
Navigation fixes

conrad-vanl:
This fixes two issues: one performance-related with swipe-to-go-back and one bug with tab bar switching on web.

More details:

1. Swipe to go back previously triggered a re-render as soon as your finger left the screen. Say you swiped more then half way and release - the component would trigger `history.goBack()`, which would update the browser URL, then complete the animation. This waits for the animation to complete before triggering navigation changes.

2. On web there was an error where if you went to a child page (ie...opened giving) then changed tabs, you'd be taken to the existing tab you were on instead of the tab you clicked on. This happened because we weren't actually updating the components state to point to the new location, we were only navigating back to the tabs screen - which was still rendering the old tab instead of the new tab.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


conrad-vanl:
@burkeshartsis @Lepozepo whoever can get to this first, thanks!

PR #300, opened 2018-02-08
Sentry

conrad-vanl:
This adds Sentry for error reporting. Currently the keys tie it to my account on expo builds, but if someone at NewSpring signs up for an account we can get it setup on yours.

Also had to change the way env vars are loaded, as they weren't being properly loaded in native. Now, we generate `app.json` at runtime, and inject `.env` values into it - which is the documented way to provide configuration data for expo.

Also, to keep from exposing all of `process.env` to a browser or react-native client device, we now only expo ENV vars that begin in `APP_`.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


src/index.js
```diff
@@ -1,19 +1,31 @@
 import { AppRegistry, Platform } from 'react-native';
 import { ApolloProvider } from 'react-apollo';
 import { nest, withProps } from 'recompose';
+
+import Sentry from '@utils/sentry';
+import Settings from '@utils/Settings';
```
IsaacHardy: Is this what we should use to pull env variables? I am assuming this util file pulls in env vars in a way that React Native can process them?

conrad-vanl: Correct

conrad-vanl:
I think theoretically we could add your "generate .env" file command to the travis deploy pipeline - `.travis.yml` - and then switch the `NODE_ENV` var in travis for different branches. Master can always deploy to alpha, we could create beta and production branches that automatically trigger builds for the respective environments as well.

Just thoughts - documenting potential approaches for down the road.

IsaacHardy:
@conrad-vanl I like that idea and is one of the reasons I built the script. Is there a `simple` way of having master be our production branch and building alpha and beta branches off of it? I would prefer to have master as, well the master branch.

conrad-vanl:
> Is there a simple way of having master be our production branch and building alpha and beta branches off of it? I would prefer to have master as, well the master branch.

Of course. Maybe the best solution would be to get away from having environment branches, and create tagged releases in git, and a tagged release would trigger a beta and/or production build. Can think through this some more.

But essentially - I would think a merge to master shouldn't necessarily trigger a new app-store release. The process I'm envisioning is a merge to master deploys a new "nightly" build to expo immediately. At some point in time, you want to take master and publish it on production web, the google play store, and the app store. This could be done on a weekly interval (scheduled) or after specific milestones. How to trigger that is the question. Creating a tag is probably a really good way to do it, and falls in line with open source standards nicely.

When I get time, let me think of it some more.

conrad-vanl:
Okay @IsaacHardy here's what I think would be nice:

`master` continues to operate like it does currently: any commit to master triggers a build with the alpha environment settings (alpha rock) and publishes on Expo. It also publishes to the `gh-pages` branch, which is our hosted storybook (and eventually documentation).

You can create a release in github https://help.github.com/articles/creating-releases/
We can configure travis to handle builds for releases differently by tapping into the conditional logic here - https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A

When building for a tagged release, we can have travis prepare testflight and google play beta builds, and continue to follow the process you're outlining here:

> create a Beta app for iOS and Android, all the while allow the build to be contained in the manual process of promoting it to testflight

However, we still have separate `beta` and `production` environment settings - such as the beta rock instance vs the production rock instance. We wouldn't want to upload a `beta` env app to testflight, and promote it to production, as the app itself would still be pointed at the `beta` environment. We'd need to rebuild with `production` settings.

So I think if we wanted to create a build pointing at `beta` environment settings, but still have it published to testflight, we can tap into the "pre-release" option when creating the github release. If the release is a "pre-release", we can configure travis to use `beta` environment settings. We can use `-beta` (or `RC`) in the versioning as well to denote that.

You would then just have to create a "real" release pointing at production settings and not promote the beta version in testflight / google play.

Thoughts?

conrad-vanl:
Or, have travis not upload builds to testflight / google play _ever_ that our pointing at anything other then the production env settings, and just use expo app for alpha/beta builds.

Can still use tagged releases for beta env builds, travis just wouldn't prepare testflight / google play builds for beta builds, and just send them to expo instead.

IsaacHardy:
I like that last bit you left there. Are there any inefficiencies in testing a beta build through expo instead of native app stores?

IsaacHardy:
@conrad-vanl FWIW, my philosophy is simplicity and adoptability. I want NewSpring's solution to work agnostically across multiple church needs.

conrad-vanl:
> Are there any inefficiencies in testing a beta build through expo instead of native app stores?

Not that I'm aware of

> @conrad-vanl FWIW, my philosophy is simplicity and adoptability

Might be worth considering dropping `beta` and `alpha` environments then, and just have a single `staging` and `production` environments (aka pre-release and release)

PR #321, opened 2018-02-12
add `KeyboardAwareScrollView`s

### LINKED ISSUE #282: KeyboardAvoidingView usage
Make sure we're properly implementing `KeyboardAvoidingView` and `KeyboardAvoidingScrollView`

conrad-vanl:
Fixes #282 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [x] Review new test logic


conrad-vanl:
@burkeshartsis I made this better. It's not perfect, but I think it should be putting most inputs close to in-view now. It's an improvement over current master, that's for sure :-D

Might need to go back to drawing board long-term in figuring out how better to support keyboard in android

src/@ui/KeyboardAwareScrollView.js
```diff
@@ -0,0 +1,8 @@
+import { defaultProps } from 'recompose';
+import { Platform } from 'react-native';
+import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view';
+
+export default defaultProps({
+  enableOnAndroid: true,
```
burkeshartsis: It looks like the original approach in this PR was to import 'KeyboardAwareScrollView' from 'react-native-keyboard-aware-scroll-view` where needed. Since then there was a bug with Android that needed to be addressed across the board that made creating this wrapper more necessary. However, we are still passing `enableOnAndroid` everywhere else in the app even though it's technically no longer needed. It would be nice to clean up those other references in light of this wrapper component handling it by default.

burkeshartsis: addressed in aaac57c

PR #325, opened 2018-02-13
supposedly add support for iOS AppLinks

conrad-vanl:
__NOTE:__ Read [this comment](https://github.com/NewSpring/Apollos/pull/325#issuecomment-377281736) below for synopsis on this PR

----

This _might_ add support for iOS AppLinks, but there's no way to test until there's a testflight build up. 💯 

(or maybe if we create some sort of expo build and open with xcode...)

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


src/@ui/NativeWebRouter/DeepLinking.js
```diff
@@ -24,6 +27,7 @@ class DeepLinking extends Component {
   };
 
   async componentDidMount() {
+    Linking.getInitialURL().then(this.handleChange);
```
Lepozepo: Mmm, this is going to push a new entry into history for all clients, including the one's that aren't coming from web (clients will be able to go back to the same thing on first load I think). Not too sure if this should be there.

conrad-vanl: @Lepozepo the `push` function below only pushes a path if there is a path to push. Expo docs show that if the app is _booted_ from scratch it won't fire a `url` event, it will just populate `getInitialUrl()`. So the use case of this line of code is clicking an app link that opens the app after it's been closed, vs when its already opened.

conrad-vanl:
@burkeshartsis this doesn't appear to change the way giving flow works.

@IsaacHardy I can't seem to test this locally - the expo standalone builds that I can build with my expo account in the simulator don't seem to respond to app links. I think I need to use the same expo login you use to build the testflight/play store builds. Which account are you using for standalone builds?

conrad-vanl:
Okay!! I am re marking this ready for review. Here's the new-new net changes for this PR:

1. Regular deep link support in both iOS and Android
2. Firebase dynamic link support in iOS
3. Search results open in-app versus kicking out to browser (falls back to in-app browser if we can't show that url in-app)

The fun part: there is no way to test dynamic links without creating a TestFlight build. However, you can test deep links! To test deep links when running expo locally, you need to know your package server's IP address (or, you can use `localhost` if you're on the simulator). For example, my computer is `192.168.1.40`. I can then go to the device's web browser, and type in links like this:

`exp://192.168.1.40:19000/+/series`
`exp://192.168.1.40:19000/+/studies`

and so on...any url path that our router supports will open up with that link type (make sure to include "+" at beginning).

For universal app link / dynamic link support on iOS, you can reference the code changes in `AppRouter.js`. Essentially, the way these work is when the app opens from a dynamic link, we grab the link that was used (for ex: `http://rm2y5.app.goo.gl/t8NBj7ht1CeClF8z1`). We `fetch` that link to see where that link is supposed to go on the website (that ex goes to: `https://newspring.cc/sermons`), and then we run that URL through our fancy new `getAppPathForUrl` util function that I ~stole~ took from Holtzman. If we have an app-path for the given URL, we route the user to that path. "

You can test this flow on search results - as now search results use the same `getAppPathForUrl` util to map search results URLs to in app links. If we can open a search result in-app, we now navigate the user to it in-app. If not, we push out to an in-app browser.

Here's a llama:

image

richarddubay:
I want to pass this PR _just_ because of the llama.

burkeshartsis:
llama perfection.

PR #333, opened 2018-02-13
Remove run-with-settings from web

conrad-vanl:
`run-with-settings` no longer needed on web since `env.js` was patched to properly consider `NODE_PATH` being set.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


conrad-vanl:
@burkeshartsis ignore the exponent failure, that's the same "not logged in" error we get intermittently. Netlify build passed, which means this PR worked. If PR build passes, this is good to merge!

PR #338, opened 2018-02-14
Sermon audio

### LINKED ISSUE #328: Sermon audio
Sermons need "Listen to Audio" button

conrad-vanl:
Fixes #328

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [x] Review new test logic


conrad-vanl:
@burkeshartsis I got the first part of this done - setting up the mediaPlayer store to play arbitrary tracks. Now we just need the UI in sermons that fires the action. Here's how you'd do it:

Use the `withNowPlaying` HOC:

```
import { withMediaPlayerActions } from '@data/mediaPlayer'
```

Which adds a function in to the component's prop `setNowPlaying`, which looks like this:

```
this.props.setNowPlaying({
  currentTrack: {
    title: PropTypes.string,
    file: PropTypes.string,
    duration: PropTypes.string,
  },
  playlist: {
    title: PropTypes.string.isRequired,
    images: ConnectedImage.propTypes.source.isRequired,
    colors: PropTypes.arrayOf(PropTypes.shape({ value: PropTypes.string })).isRequired,
    tracks: PropTypes.arrayOf(track), // optional: omit if there's only the current track
  },
});
```

IsaacHardy:
How is the progress coming on this PR?

burkeshartsis:
@IsaacHardy it's mostly done. Need to sort out some bugs that @conrad-vanl probably wasn't able to test till we had a UI for it. Hoping to get it reviewed by EOD.

burkeshartsis:
So technically this works now. However, the media player (bottom of the screen) was designed for multiple track albums as that was it's original use case. Sermon audio (and something we maybe haven't thought completely thought but a worship single) is only a single track. To really complete this PR we need to update the media player to no show player controls that aren't relevant to a playlist with only 1 track (i.e. next, previous, shuffle).

conrad-vanl:
You might look and see what the current app does. IRC, it still shows those player controls

burkeshartsis:
Another issue I'm seeing is that the apps seems to get super unresponsive once sermon audio has been triggered. Wondering if there is some runaway code somewhere.

conrad-vanl:
> Another issue I'm seeing is that the apps seems to get super unresponsive once sermon audio has been triggered. Wondering if there is some runaway code somewhere.

Try in production mode. I've noticed this issue in dev a few times, but haven't been able to reproduce consistently and in production.

burkeshartsis:
This isn't a problem currently because we redirect web traffic but if we didn't web would be broken as the `id` string [here](https://github.com/NewSpring/Apollos/blob/sermon-audio/src/music/Player/index.js#L174) is undefined for sermon audio. @conrad-vanl thoughts.

burkeshartsis:
This error appears when you select the more (...) overflow in the fullscreen audio player 😢

PR #340, opened 2018-02-14
polish touch feedback on iOS

### LINKED ISSUE #306: Links flash white on ios when thumbing/scrolling through app.
## Acceptance Criteria

- [ ] Upon scrolling in app, when touching different sections(links) default white "flash" should not occur.

## Context

Currently, when you thumb through the app on ios, anytime that you touch a link, the whole element flashes white.

conrad-vanl:
Fixes #306 by adding a very small delay in touch feedback, which prevents touch feedback from triggering as you're scrolling a feed. Also lowers the opacity affect on touch a bit so that the touch feedback isn't as drastic.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [x] Review new test logic


src/@ui/Touchable/index.js
```diff
@@ -1,5 +1,11 @@
 import { Platform, TouchableNativeFeedback, TouchableOpacity } from 'react-native';
+import { defaultProps } from 'recompose';
 
-const Touchable = Platform.OS === 'android' ? TouchableNativeFeedback : TouchableOpacity;
+const IOSTouchable = defaultProps({
+  delayPressIn: 50,
```
burkeshartsis: @conrad-vanl In stead of making this default could we just pass it in on touchables in feeds like cards?

conrad-vanl: We could... granted, i'd say that there are more touchables in scrollviews then there are touchables not in scrollviews. I'd almost rather see about how low we can lower the delay value and still keep scrollviews from firing touch effects on scrolling.

burkeshartsis: Yeah I looked into this a little bit to see if there might be a better way to address this given that delaying user interaction is usually the last resort approach. Unfortunately, the way react native works by "handing off" the touch event from components within the `ScrollView` to the `ScrollView` itself on drag is pretty hard to work around leaving no other options than to add a delay like you have. I might play look at playing with the `delayPressIn` value a bit if I get a chance.

conrad-vanl:
@burkeshartsis in that case, lets just use `TouchableWithoutFeedback` for cards for now. That's where the main issue is. We can pass a `component` prop into the wrapping `<Link>`

PR #352, opened 2018-02-19
fix empty array error and implement basic aspect ratio sorting

### LINKED ISSUE #349: Look into image cropping on content cards
> all the images for the devos and studies and maybe also sermon entries don't looked scaled right either

I believe the reference here is that Card images are being cropped - ie...`ConnectedImage` isn't showing the square image. This could be solved by tweaking the graphql queries, sorting the images better, or something else.

conrad-vanl:
Fixes #349 as well as an issue affecting current master builds where if a content item had no images, would result in a crashed app.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [x] Review new test logic

### DIFF
--- a/src/@utils/content/getItemImages.js
+++ b/src/@utils/content/getItemImages.js
@@ -4,7 +4,17 @@ const getItemImages = (item) => {
   let images = item.photo || get(item, 'content.images', []);
   if (!images.length) images = get(item, 'parent.content.images', [{}]);
   if (typeof images === 'string') return images;
-  return images[0].url; // TODO short circuit. Highliner should handle image selection in the array
+  // Sort 1:1 images to the front of array.
+  // FWIW, `ConnectedImage` will sort out which image is best to use from a resolution
+  // perspective, which is why we just kind of blindly push 1:1 images to the front,
+  // as it doesn't consider aspect ratio, just resolution.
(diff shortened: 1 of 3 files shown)


src/@utils/content/getItemImages.js
```diff
@@ -4,7 +4,17 @@ const getItemImages = (item) => {
...
+  // FWIW, `ConnectedImage` will sort out which image is best to use from a resolution
+  // perspective, which is why we just kind of blindly push 1:1 images to the front,
+  // as it doesn't consider aspect ratio, just resolution.
+  images = [].concat(images).sort((a, b) => { // make sort pure
+    if (a && b && a.fileLabel === b.fileLabel) return 0;
+    if (a && a.fileLabel === '1:1') return -1;
+    if (b && b.fileLabel === '1:1') return 1;
+    return 0;
+  });
+  return images;
```
conrad-vanl: Basically - the error on master was the short circuit we had originally implemented to just return the first image url. This is no longer needed as `ConnectedImage` now supports an opaque data type (array, object, etc) instead of just a url string. 

Secondly, added this sort function so that we prioritize displaying of square images (not all content items have square images so we have to ask for both from heighliner). This fixes #349

PR #377, opened 2018-03-01
374 giving history

### LINKED ISSUE #374: Giving History Issues
## Acceptance Criteria

- [ ] It should show a loading indicator when the user scrolls down to fetch more data.
- [ ] It should keep the giving history in the correct order after fetching.

## Context

Given that a user wants to see more of their giving history, when they scroll down to see more, then a loading indicator should display on the screen when the app is fetching more data.
Given that a user wants to see their giving history in the correct reverse chronological order, when the users has "fetched more", then they should still see their transactions in the correct order.

Attaching two items here. One is an animated GIF of the "no loading indicator" problem. The other is an image of the top of the giving history feed AFTER having fetched more of my history.

No Loading Indicator

no-loading-indicator

Top of Giving History After Fetch

messed-up-giving-hitory

IsaacHardy:
Fixes #374.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/src/give/ContributionHistory/ContributionHistoryList.js
+++ b/src/give/ContributionHistory/ContributionHistoryList.js
@@ -1,9 +1,7 @@
 import React, { PureComponent } from 'react';
-import {
-  View,
-} from 'react-native';
+import { View } from 'react-native';
 import PropTypes from 'prop-types';
-import { groupBy, map } from 'lodash';
+import { groupBy, map, orderBy } from 'lodash';
 import HistoricalContributionCard from '@ui/HistoricalContributionCard';
 import FlatList from '@ui/WebCompatibleFlatList';
 import FlexedView from '@ui/FlexedView';
@@ -22,22 +20,21 @@ class ContributionHistoryList extends PureComponent {
     fetchMore: PropTypes.func,
     isLoading: PropTypes.bool,
     refetch: PropTypes.func,
-    transactions: PropTypes.arrayOf(PropTypes.shape({
-      date: PropTypes.string,
-      person: PropTypes.shape({
-        firstName: PropTypes.string,
-        lastName: PropTypes.string,
-        photo: PropTypes.string,
+    transactions: PropTypes.arrayOf(
+      PropTypes.shape({
+        date: PropTypes.string,
+        person: PropTypes.shape({
+          firstName: PropTypes.string,
+          lastName: PropTypes.string,
+          photo: PropTypes.string,
+        }),
+        id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+        amount: PropTypes.number,
+        account: PropTypes.shape({
+          name: PropTypes.string,
+        }),
       }),
-      id: PropTypes.oneOfType([
-        PropTypes.string,
-        PropTypes.number,
-      ]),
-      amount: PropTypes.number,
-      account: PropTypes.shape({
-        name: PropTypes.string,
-      }),
-    })),
+    ),
     onPressNoDataButton: PropTypes.func,
     onPressContributionCard: PropTypes.func,
     FilterComponent: PropTypes.any, // eslint-disable-line
@@ -59,7 +56,9 @@ class ContributionHistoryList extends PureComponent {
       {item.transactions.map(transaction => (
         <Touchable
           key={transaction.id}
-          onPress={() => { this.props.onPressContributionCard(transaction.transactionId); }}
+          onPress={() => {
+            this.props.onPressContributionCard(transaction.transactionId);
+          }}
         >
           <HistoricalContributionCard
             amount={transaction.amount}
@@ -82,7 +81,14 @@ class ContributionHistoryList extends PureComponent {
       );
     }
 
-    const transactionsPerYear = map(groupBy(this.props.transactions, 'year'), (transactions, year) => ({ year, transactions }));
+    const transactionsPerYear = orderBy(
+      map(groupBy(this.props.transactions, 'year'), (transactions, year) => ({
+        year,
+        transactions,
+      })),
+      ['year'],
+      ['desc'],
+    );
     let Header = (
       <View>
         <this.props.FilterComponent />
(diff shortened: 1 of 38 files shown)


IsaacHardy:
It seems there are still some odd cases taking place on web that we'll need to investigate later, however, the current functionality is now adequate for this bug's scope.

IsaacHardy:
Native on iOS

IsaacHardy:
Web

src/give/ContributionHistory/ContributionHistoryList.js
```diff
@@ -82,7 +81,14 @@ class ContributionHistoryList extends PureComponent {
       );
     }
 
-    const transactionsPerYear = map(groupBy(this.props.transactions, 'year'), (transactions, year) => ({ year, transactions }));
+    const transactionsPerYear = orderBy(
```
burkeshartsis: Could this be better handled in Highliner directly? Thoughts?

IsaacHardy: I think it could, but right now that might be creating more scope than we can accept (since we've already added five new stories to scope). I want the best solution for this, however I do not know the best answer as of now.

burkeshartsis: Will handling this client side create issues when `fetchMore` is called for additional giving history that will come down unsorted?

IsaacHardy: @burkeshartsis I can confirm through my local testing that fetchMore does not cause issues with this fix as seen in the GIF above. However, my testing was at a high level.

Lepozepo: @IsaacHardy ordering on the server would definitely be ideal, let me have a quick look at the server

IsaacHardy: @Lepozepo Will the ordering need to take place here: https://github.com/NewSpring/Heighliner/blob/master/src/rock/models/finances/models/Transaction.js#L362 ?

Lepozepo: @IsaacHardy it seems to be running this function to resolve transactions: https://github.com/NewSpring/Heighliner/blob/master/src/rock/models/finances/models/Transaction.js#L185 which seems to be ordering already. I'm reviewing the query, there's some nonsense going on there especially in the `.slice` part.

Lepozepo: Yeah, and I think we're getting dupes too

IsaacHardy: @Lepozepo That would explain some odd behavior when I first started looking into this issue.

Lepozepo: Ok, the silliness on the server is patched. Just need to fix tests on that branch. The duplicate keys issue is rooted in the way rerenders are being triggered. Somehow it's causing `fetchMoreResolver`'s returned function to run multiple times on the same results (which in turn updates apollo's store and pushes the same results onto it)

IsaacHardy: @Lepozepo Do you think this is happening more than just one play within Heighliner?

Lepozepo: It could be happening in other resolvers as the bug was more in the lack of `attributes` from which the query could order from. So I'd look at all the queries using `attributes` and make sure the array contains all used columns. The way the `cache` works right now also seems to cause silent errors if the query fails (which is what made the issue so difficult to identify to begin with). The push I just made is making sure `fetchMoreResolver` does not retrigger/resolve prematurely to protect it against active inefficiencies in render lifecycles, this resolves the root of the duplicate ID's issue (it may have been affecting multiple lists across the app too)

Lepozepo:
newspring

Lepozepo:
This could use some UI love in the form of showing an indicator for when a refetch is underway. @burkeshartsis I'm not sure if FlatList has something specific for that already baked in, if not, you can leverage `onRefresh` and `onEndReached`.

Lepozepo:
Depends on [this](https://github.com/NewSpring/Heighliner/pull/373)

burkeshartsis:
@Lepozepo we just has to hook up `onRefresh` in `FlatList`

IsaacHardy:
@burkeshartsis Could you direct me on where I should investigate to implement this UI update?

burkeshartsis:
@IsaacHardy Absolutely! I believe [this example](https://github.com/NewSpring/Apollos/blob/master/src/%40ui/FeedView/index.js#L90) should suffice. You just have to make sure Apollo is sending `isLoading` down.

IsaacHardy:
@burkeshartsis @Lepozepo It seems that the ordering is still off. It starts with 2018 at the top, but then reverses the transactions once you scroll down. Creating a GIF to demonstrate.

Lepozepo:
@IsaacHardy maybe the `groupBy` is messing up the order and we do need to reinstate the `orderBy` that you had set up. This, however, is one situation in which it would be beneficial to be mindful of performance as we have more data points, javascript will take longer to order/group. We could potentially use the transducer pattern (which will ensure we at least only do one loop on the array) or even better we could try to extend `fetchMoreResolver` to take a sort of `reducer` so that it only maps new results into each year.

IsaacHardy:
@Lepozepo From your current understanding, how complex is that last options. That might be the best solution.

IsaacHardy:
Found this within the docs. Could be useful. https://www.apollographql.com/docs/react/features/caching.html#resultReducers

IsaacHardy:
@Lepozepo After doing some research, it seems we will have to update the `fetchMoreResolver` to generate a subscription to the server which in turn would allow us to create a reducer for this query. 

To create a subscription we would have to update the server as well. Am I on the right path or did you see this in a different way?

Lepozepo:
@IsaacHardy I've looked over `groupBy` and `fetchMoreResolver`. `fetchMoreResolver` is doing it's job correctly now, no need to update that and `groupBy` does reverse the order of the list (you can run `_.groupBy([{year: 2018}, {year: 2017}], 'year')` to see for yourself ^_^ and [here](https://github.com/lodash/lodash/blob/master/groupBy.js#L28) is the culprit, maybe `unshift` would respect order). The array generated by `groupBy` isn't huge though so we should be ok with `orderBy` in there.

Lepozepo:
@burkeshartsis this should be good to merge now

IsaacHardy:
@Lepozepo Appreciate the work!

PR #385, opened 2018-03-06
369 Netlify toml

### LINKED ISSUE #369: Deploy Apollos 6.0.0 to Prod
## Acceptance Criteria

- [ ] Apollos 6.0.0 should be deployed to NewSpring Production environment.

## Context

We need to research and construct a solution for deployment of Apollos 6.0.0 on the web.

IsaacHardy:
First part for fixing #369.

* This updates the script to build web to run with different node environments
* It creates a netlify.toml file to run different commands based on a specific branch. 
* It updates the `create-env-files` script to let us use `.env` file instead of keeping that information on Netlify. This should allow us to set certain branches up to be "beta" and "alpha" with specific env variables for each all the while keeping this data closed from the public. 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [x] Review new test logic

### DIFF
--- a/.scripts/generate-env-files.js
+++ b/.scripts/generate-env-files.js
@@ -2,15 +2,16 @@ const Path = require('path');
 const FS = require('fs');
 const fetch = require('node-fetch');
 const TOKEN = process.env.RAW_SETTINGS_TOKEN;
+const env = process.env.NODE_ENV;
 
 (async () => {
   try {
     const response = await fetch(
       'https://raw.githubusercontent.com/NewSpring/ops-settings/master/sites/apollos/settings.json' +
-      '?token=' +
-      TOKEN,
+        '?token=' +
+        TOKEN,
     );
-    
+
     const settings = await response.json();
 
     const localContent = `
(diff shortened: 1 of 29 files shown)


.scripts/generate-env-files.js
```diff
@@ -40,20 +41,14 @@ APP_HEIGHLINER_URL='${settings.prod.heighliner}'
...
-      if (err) console.log(err);
-      console.log("Local environment settings created.");
-    })
-    await FS.appendFile(Path.resolve(__dirname, '../.env.test'), testContent, (err) => {
-      if (err) console.log(err);
-      console.log("Test environment settings created.");
-    })
-    await FS.appendFile(Path.resolve(__dirname, '../.env.production'), productionContent, (err) => {
+    const envContent =
+      env === 'production' ? productionContent : env === 'test' ? testContent : localContent;
```
burkeshartsis: Is there any way to avoid this nested ternary 🤮

netlify.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[build]
+# This is the directory that you are publishing from (relative to root of your repo)
+  publish = "build/"
+# default build command
+  command = "npm run build:alpha"
+
+# Production context: All deploys from the Production branch set in your site's deploy settings will inherit these settings.
+[context.production]
+  command = "npm run build"
```
burkeshartsis: let's point all of these at alpha for the time being so nothing breaks.

.scripts/generate-env-files.js
IsaacHardy: 😆 I could go the ole' fashion way, but didn't think it was that terrible with only three options.

IsaacHardy:
@burkeshartsis Updated with new [context.production] command.

.scripts/generate-env-files.js
burkeshartsis: 🙃

IsaacHardy: hahahahaha

PR #398, opened 2018-03-12
312 closing media player

### LINKED ISSUE #312: Closing of music player
## Acceptance Criteria

- [ ] can't quit song from bottom of screen when navigating in the app (it does allow you to pause)

@IsaacHardy / @FrankGrand this was reported by Michael. I don't believe this is possible in the current app either? but ideas on how best to implement it? Simple might be to put like a X second timeout that if you stop playing music, after so many seconds, the player closes.

burkeshartsis:
Fixes #312 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [x] Review new test logic


burkeshartsis:
Ok this is mostly done. Just need a function passed down to the thumbnail that will pause the music and remove/unmount the miniplayer from the react tree. The only way I know how to currently do that would be to pass a function from the miniplayer's parent down that sets state. Then conditionally render the miniplayer based on that state. @Lepozepo thoughts/better idea?

Lepozepo:
@burkeshartsis I think the logic behind rendering the controls needs to be in local state at the `Player` component considering the decisions we've made so far throughout the app. Line 178 in `/Player/index.js` should pass this closing function down all the way to mini controls or make it available via context (`Player` is an ecosystem of components anyway so context makes sense here too).

burkeshartsis:
Because the media player and most notably the mini controls live in a cardstack in order to really dismiss the media player we are going to need to push a new or pop the old route. @conrad-vanl do you have an idea how I could replace the touchable `onPress` within `MiniControlsThumbnail` with a `Link` that pops to the correct route (removes the media player controls entirely)?

conrad-vanl:
> controls live in a cardstack in order to really dismiss the media player we are going to need to push a new or pop the old route

I don’t believe this is true. You should be able to call one of the Apollo mutations and clear the media player state and the player will go away

IsaacHardy:
@burkeshartsis Cannot seem to get travis to build the branch.  Working on a solution so we can merge this in!

burkeshartsis:
Ahh! I think know what's wrong @IsaacHardy. I'll look into it now.

IsaacHardy:
@burkeshartsis Everything looks good here. Are we good to merge?

burkeshartsis:
@IsaacHardy Tests are only passing because I rolled back the Jest version and commented out the test that was failing. Just wanted to rule out some factors that way (travis). Now we have a path forward to try and asses why the test wont pass.

burkeshartsis:
This is the line that is killing Jest. Not sure why but getting closer.

[`backgroundColor: Color(theme.colors.background.default).fade(theme.alpha.low),`](https://github.com/NewSpring/Apollos/blob/312-closing-media-player/src/music/Player/MiniControls/MiniControlsThumbnail/index.js#L31)

burkeshartsis:
All tests are now passing

PR #411, opened 2018-03-19
408 - Refresh queries after giving

### LINKED ISSUE #408: App data does not update until a reload of the main giving component
## Acceptance Criteria

- [ ] App data should update after the initial action is taken.

## Context

In certain scenarios, the app data does not update after an action has been made. For example, once you return to the dashboard after giving a specific amount, the contribution will not be present until you navigate back to the giving component from the section menu. 

Giving: 

Schedule:

richarddubay:
Fixes #408 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


richarddubay:
@burkeshartsis @conrad-vanl @IsaacHardy This is really more for discussion at this point I think. I'm not certain about what is going on here. Initially, I thought I could just remove the `hasQueryParams` check around the `setPaymentResult` call that re-fetches the queries, but it turns out that the `setPaymentResult` function isn't really happy with "undefined" success and error params. The gift would still go through, but it showed the "Uh Oh!" screen. So I had to set those when `hasQueryParams` is false. 

So...this seems to work. When I go back to the Dashboard after giving the updated data is there. Is there a better way to do this? What am I missing?

conrad-vanl:
Lets let @Lepozepo review this

Lepozepo:
Ah, I see the problem. The issue is specific to Android and web (because both finish checkout within the same "session"). The logic on `PaymentComplete` in `master` is right, the one that's wrong is the one in `PaymentConfirmationForm` that is also calling `setPaymentResult` but without the refetch flag set to `true`. I don't think refetching should be optional for this function so I'm removing that (the dev should not need to know to refetch)

Lepozepo:
Also this PR was trying to `JSON.parse` `error` which returns an `error.message` string and cannot be parsed (causing the app to crash). Let me double check that is not the case anymore :D

Lepozepo:
Mmm, there may be a deeper issue here though. The server is responding correctly but the array the component is catching is **sometimes** doesn't have the new transaction (almost like the client is returning a cached response if you browse away from the success page too quickly). Looking through it.

Lepozepo:
I think the deeper problem is we can't `await` for `refetch` to finish as described [here](https://github.com/apollographql/apollo-client/issues/1618) and it seems subsequent re-renders are messing up the internal state in the `graphql` decorators instance even though the network is returning the correct results. The temporary fix could be setting a pollinterval to ensure results are always fresh but that's pretty inefficient on the server. I'll try recalling setResults again on the confirm screen, it seems that was obscurely fixing the issue ^_^.

Lepozepo:
:'( nvm, the issue can still occur. I'll set the query to poll instead.

Lepozepo:
Should be good to go now but I think there's more we can do to make this more efficient later on.

newspring

Lepozepo:
@conrad-vanl let me know what you think ^_^

IsaacHardy:
@conrad-vanl That sounds good. I cannot think of a better place to poll.

PR #438, opened 2018-03-27
Update env script Token

IsaacHardy:
We now use a permanent token.


IsaacHardy:
@conrad-vanl Feel free to merge this in if everything looks good on your end.

IsaacHardy:
@conrad-vanl Something weird is happening with the travis build.

PR #445, opened 2018-04-02
Fix DeepLinking white screen on app load

conrad-vanl:
Fix an issue where deep linking was being a little too proactive and routing to empty page

Also updates `yarn.lock`

Fixes [issue-number]

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/src/@ui/NativeWebRouter/DeepLinking.js
+++ b/src/@ui/NativeWebRouter/DeepLinking.js
@@ -51,8 +51,8 @@ class DeepLinking extends Component {
       pathname = pathname.substr(2);
     }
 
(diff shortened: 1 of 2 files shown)


src/@ui/NativeWebRouter/DeepLinking.js
```diff
@@ -41,7 +42,14 @@ class DeepLinking extends Component {
   };
 
   push = async (url = '') => {
-    const pathname = url.replace(this.baseUrl, '');
```
conrad-vanl: Before, `baseUrl` would be equal to something like `exp://localhost:9000` or `exp://localhost:9000/+/series`

PR #453, opened 2018-04-03
Fix issue with universal link loading

conrad-vanl:
Fixes an issue with universal link loading on mount. Essentially, the
`DeepLinking` component was re-mounting when `AppRouter` was
re-rendering, causing the app to enter into a recursive cycle showing
an endless loading indicator.

Fixes [issue-number]

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "eslint-plugin-jest": "21.2.0",
     "eslint-plugin-jsx-a11y": "6.0.2",
     "eslint-plugin-react": "7.4.0",
-    "exp": "^50.0.5",
+    "exp": "^51.4.0",
     "jest": "^21.2.1",
     "jest-expo": "^25.1.0",
     "react-native-scripts": "^1.6.0",
--- a/package.json
+++ b/package.json
@@ -36,28 +36,19 @@
   },
   "main": "./node_modules/react-native-scripts/build/bin/crna-entry.js",
   "scripts": {
-    "start":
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "eslint-plugin-jest": "21.2.0",
     "eslint-plugin-jsx-a11y": "6.0.2",
     "eslint-plugin-react": "7.4.0",
-    "exp": "^50.0.5",
+    "exp": "^51.4.0",
```
conrad-vanl: bump exp client for builds (not the expo sdk)

src/@data/Client/errorLink.js
```diff
@@ -0,0 +1,27 @@
+import { onError } from 'apollo-link-error';
+import sentry from '@utils/sentry';
+
+const link = onError(({
```
conrad-vanl: getting mysterious network errors and want to start capture them in sentry to see what's going on

conrad-vanl:
@IsaacHardy this PR should be ready to go now, and is already published in the latest tesflight/play store builds. FWIW, universal links are still not working properly on store builds, but they are not causing errors now. At the current moment, I believe there is an expo bug preventing us from fully implementing universal linking. I will look into this more after launch. Essentially, if you click on an "Open in App" link on the website - it will take you to the app - but it won't open the correct content currently. The code in this PR _does_ work, but expo isn't passing the linking url in so we have no idea where the user came from.

PR #462, opened 2018-04-05
Performance: Splash Screen

conrad-vanl:
This triggers a feature on ios that should improve loading times. Android has this feature enabled by default.

Also, @IsaacHardy ... could the color issue just have been "fixed", and splash screen backgroundColor can be the same between ios and android now? Not sure...

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/app-config.js
+++ b/app-config.js
@@ -24,8 +24,9 @@ module.exports = env => ({
       ],
       icon: 'src/assets/icons/ios/icon-1024x1024.png',
       splash: {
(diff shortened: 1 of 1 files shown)


IsaacHardy:
@conrad-vanl So weird, but there seems to be two stages to the splash screen load now. Before, the 2nd half would be off-color, now it's the first. I think this is good for now to be honest, because its really hard to catch unless you're looking for it:

PR #481, opened 2018-04-09
fix for splash screen weirdness

IsaacHardy:
There is some weird stuff happening with the splash screen lately, and I believe this will fix the issue.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [x] Review new test logic


IsaacHardy:
@richarddubay I pushed new code. Should work on Android now.

PR #510, opened 2018-04-16
Fix restored checkout in giving

conrad-vanl:
Alright, so in #501 we downgraded the `apollo-client` as the newest version of the apollo-client introduces a bug in android (documented here: https://github.com/apollographql/apollo-client/issues/3236). We originally upgraded to apollo-client@2.2.8 because the version we were on, 2.0.x, had problems in error handling in network requests, causing frequent crashing on Android. 

So, to fix, I upgraded `apollo-client` (and `react-apollo`) back to the newest versions, and added a polyfill for `setPrototypeOf`, just for Android.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


IsaacHardy:
Lookin' good.

MichelDiz:
This Merge extends tod Apollo-Boost?

PR #511, opened 2018-04-16
Stability fixes

conrad-vanl:
Started looking at stability stuff I could find before getting wind of the Restored Checkout issue on iOS. Here are two things I came across:

1. We are still querying for `embedUrl` in the live now query. This is no longer needed with CHOP, so I removed. Dramatically speeds up these queries and I'm hoping will fix these random 503 errors I'm seeing

2. After the universal-link fixing in android, the featured cards in the discover tab are awkward, as they start to open a web browser, then bounce you back to the app. So we go ahead and just open them in-app from the get go.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


IsaacHardy:
Looks like we had a bunch of dirty code. Good job.

PR #515, opened 2018-04-18
Add restricted paths and use singleInstance for universal linking on android

conrad-vanl:
Previously Android app was responding to _any_ url that starts with `newspring.cc/`. This adds in a bunch of restrictions so it only responds to certain paths, so it fixes the issue @FrankGrand discovered where location cards weren't working.

Also, it switches the activity type to `singleInstance`, which _should_ also keep only one instance of the app open ever.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -51,7 +51,7 @@
     <activity
         android:name=".LauncherActivity"
         android:exported="true"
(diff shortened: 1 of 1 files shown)


IsaacHardy:
@conrad-vanl I believe this is something we'll have to test in the wild. Cannot test locally. So I approved. 😬

PR #521, opened 2018-04-19
Fix dup bad cc bug

conrad-vanl:
So there were two issues to solve.

First, the bad credit card actually _wasn't_ being added, we just weren't properly showing an error message.

Second - and the real bug - we were saving some of the state from saving a credit card so if the user tried again, they wouldn't have to re-enter all their information again. When checking out with a valid credit card, the checkout process was reacting to state that said "Save Credit Card with this name" - except the name was the unsuccessful BAD credit card. So this is fixed by explicitly resetting state when starting a _new_ giving transaction.

I made a few other tweaks though:
- Added "Try again" buttons to both failure state screens on giving and adding payment method
- The same bug (but less severe) existed if you went to gave, added a fake CC, then selected a saved credit card to give with. This is fixed by selectively resetting some state whenever you select a new credit card
- A few other minor UI tweaks (like an Edit button on the add payment review screen)

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


richarddubay:
@conrad-vanl Since @Lepozepo already approved this, I'll go ahead and merge. :feelsgood:

PR #527, opened 2018-04-23
Fix Schedule Card for Multiple Funds

richarddubay:
This is a fix for the schedule card (in particular on the giving dashboard) so that if a schedule has multiple funds then the total of those funds appear correctly on the card.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

This is a GIF of the before behavior:

This is the "after" of the dashboard with the correct amount showing:
screenshot 2018-04-23 15 32 28

### DIFF
--- a/src/give/Dashboard.js
+++ b/src/give/Dashboard.js
@@ -1,9 +1,6 @@
 /* eslint-disable no-underscore-dangle */
 import React, { PureComponent } from 'react';
-import {
-  ScrollView,
-  TouchableWithoutFeedback,
-} from 'react-native';
+import { ScrollView, TouchableWithoutFeedback } from 'react-native';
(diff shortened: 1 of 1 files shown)


src/give/Dashboard.js
```diff
@@ -103,10 +101,14 @@ export class Dashboard extends PureComponent {
             <ScheduleCard
               key={scheduledTransaction.id}
               accountName={get(scheduledTransaction, 'details.0.account.name')}
-              amount={get(scheduledTransaction, 'details.0.amount')}
+              amount={get(scheduledTransaction, 'details')
+                .reduce((i, { amount = 0 } = {}) => i + amount, 0)
```
conrad-vanl: good catch!

PR #558, opened 2018-04-30
412 dollar sign

### LINKED ISSUE #412: Dollar Sign is not aligned with dollar amount in Give Now
## Acceptance Criteria

- [ ] The dollar sign should align vertically with the dollar amount in the giving amount input on the give now tab.

IsaacHardy:
Fixes #412.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/android/app/src/main/java/host/exp/exponent/generated/DetachBuildConstants.java
+++ b/android/app/src/main/java/host/exp/exponent/generated/DetachBuildConstants.java
@@ -3,6 +3,6 @@ package host.exp.exponent.generated;
 // This file is auto-generated. Please don't rename!
 public class DetachBuildConstants {
 
-  public static final String DEVELOPMENT_URL = "expba3ad9e96a6043fc95726c73cb381a27://t3-pit.newspring-builds.apollos.exp.direct:80";
+  public static final String DEVELOPMENT_URL = "expba3ad9e96a6043fc95726c73cb381a27://10.0.200.79:19000";
 
 }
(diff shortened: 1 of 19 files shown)


android/app/src/main/java/host/exp/exponent/generated/DetachBuildConstants.java
```diff
@@ -3,6 +3,6 @@
 // This file is auto-generated. Please don't rename!
 public class DetachBuildConstants {
 
-  public static final String DEVELOPMENT_URL = "expba3ad9e96a6043fc95726c73cb381a27://t3-pit.newspring-builds.apollos.exp.direct:80";
+  public static final String DEVELOPMENT_URL = "expba3ad9e96a6043fc95726c73cb381a27://10.0.200.79:19000";
```
IsaacHardy: Seems like this is a side effect of me testing locally.

conrad-vanl: i'd like to not let this change end up in our branches - that's your local IP address and makes no sense to anyone else. For some reason only certain machines causes that url to change (@burkeshartsis's does as well when he's at the Di office). I haven't determined exact cause yet, but I'm sure something to do with corporate networks.

Anyways, `git update-index --assume-unchanged d android/app/src/main/java/host/exp/exponent/generated/DetachBuildConstants.java` I believe will make git not see this change (after you remove this change via a commit)

PR #562, opened 2018-04-30
Secondary Nav Loading Fix

### LINKED ISSUE #549: TypeError: null is not an object (evaluating 'e.title')
https://sentry.io/newspring-church/apollos/issues/539308490/

```
TypeError: null is not an object (evaluating 'e.title')
  at default (/Users/isaac.hardy/NewSpring/Apollos/src/@utils/content/share.js:6:16)
  at onPress (/Users/isaac.hardy/NewSpring/Apollos/src/@ui/SecondaryNav/index.js:12:12)
  at onPress (/Users/isaac.hardy/NewSpring/Apollos/src/@ui/NativeWebRouter/Link.js:59:20)
  at touchableHandlePress ([native code])
  at _performSideEffectsForTransition ([native code])
...
(23 additional frame(s) were not displayed)
```

richarddubay:
Fixes #549 

This attempts to load the secondary nav (think "like and share" nav) only AFTER the content loads. This should fix errors where there is no content to share.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/src/@ui/SecondaryNav/index.js
+++ b/src/@ui/SecondaryNav/index.js
@@ -27,14 +27,16 @@ const SecondaryNav = ({
   onBackPress,
   children,
   onBackReplace,
-}) => (
-  <TabBar>
-    {backButton ? (
-      <Link pop to={backTo} icon={backButtonIcon} onPress={onBackPress} replace={onBackReplace} />
-    ) : null}
-    {children}
-  </TabBar>
-);
+  isLoading = false,
+}) =>
+  !isLoading && (
+    <TabBar>
+      {backButton ? (
+        <Link pop to={backTo} icon={backButtonIcon} onPress={onBackPress} replace={onBackReplace} />
+      ) : null}
+      {children}
+    </TabBar>
+  );
 
 SecondaryNav.propTypes = {
   backButton: PropTypes.bool,
@@ -43,6 +45,7 @@ SecondaryNav.propTypes = {
   onBackPress: PropTypes.func,
   backTo: PropTypes.string,
   onBackReplace: PropTypes.bool,
+  isLoading: PropTypes.bool,
 };
 
 export default SecondaryNav;
--- a/src/@ui/SecondaryNav/index.js
+++ b/src/@ui/SecondaryNav/index.js
@@ -29,14 +29,14 @@ const SecondaryNav = ({
   onBackReplace,
   isLoading = false,
 }) =>
-  !isLoading && (
+  (!isLoading ? (
     <TabBar>
       {backButton ? (
         <Link pop to={backTo} icon={backButtonIcon} onPress={onBackPress} replace={onBackReplace} />
       ) : null}
       {children}
     </TabBar>
-  );
+  ) : null);
 
 SecondaryNav.propTypes = {
   backButton: PropTypes.bool,
--- a/src/articles/Single.js
+++ b/src/articles/Single.js
@@ -20,38 +20,38 @@ const enhance = compose(
 
 const ShareLink = withArticle(Share);
 
-const ArticleSingle = enhance(({
-  content: {
-    authors = [],
-    title = '',
+const ArticleSingle = enhance(
+  ({
     content: {
-      isLiked,
-      body,
-      tags,
-      ...otherContentProps
+      authors = [],
+      title = '',
+      content: {
+        isLiked, body, tags, ...otherContentProps
+      } = {},
     } = {},
-  } = { },
-  id,
-  isLoading,
-}) => (
-  <BackgroundView>
-    <Header titleText="Article" backButton />
-    <ScrollView>
-      <ContentView isLoading={isLoading} {...otherContentProps}>
-        <Title>{title}</Title>
-        <ByLine authors={authors} />
-        <HTMLView>{body}</HTMLView>
-      </ContentView>
-      { // Don't render till data is ready. Consider adding placeholder views for the content above.
+    id,
+    isLoading,
+  }) => (
+    <BackgroundView>
+      <Header titleText="Article" backButton />
+      <ScrollView>
+        <ContentView isLoading={isLoading} {...otherContentProps}>
+          <Title>{title}</Title>
+          <ByLine authors={authors} />
+          <HTMLView>{body}</HTMLView>
+        </ContentView>
+        {// Don't render till data is ready.
+        // Consider adding placeholder views for the content above.
         !isLoading && <RelatedContent tags={tags} excludedIds={[id]} />}
-    </ScrollView>
-    <MediaQuery maxWidth="md">
-      <SecondaryNav>
-        <ShareLink id={id} />
-        <Like id={id} isLiked={isLiked} />
-      </SecondaryNav>
-    </MediaQuery>
-  </BackgroundView>
-));
+      </ScrollView>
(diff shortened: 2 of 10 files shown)


conrad-vanl:
@richarddubay this PR got a little out of hand I assume to using prettifier. Could you point me to the relevant lines of code in this PR to look at?

src/series/Sermon.js
```diff
@@ -71,7 +71,7 @@ const Sermon = enhance(
         </ContentView>
         <HorizontalTileFeed content={children} isLoading={isLoading} showTileMeta />
       </ScrollView>
-      <SecondaryNav>
+      <SecondaryNav isLoading={isLoading}>
```
conrad-vanl: This is the way I'd do it 👍

src/@ui/SecondaryNav/index.js
```diff
@@ -27,14 +27,16 @@ const SecondaryNav = ({
...
+  isLoading = false,
+}) =>
+  !isLoading && (
+    <TabBar>
+      {backButton ? (
+        <Link pop to={backTo} icon={backButtonIcon} onPress={onBackPress} replace={onBackReplace} />
+      ) : null}
+      {children}
+    </TabBar>
+  );
```
richarddubay: I added a `isLoading` prop to the Secondary Nav here.

src/articles/Single.js
```diff
@@ -20,38 +20,38 @@ const enhance = compose(
...
-      <SecondaryNav>
-        <ShareLink id={id} />
-        <Like id={id} isLiked={isLiked} />
-      </SecondaryNav>
-    </MediaQuery>
-  </BackgroundView>
-));
+      </ScrollView>
+      <MediaQuery maxWidth="md">
+        <SecondaryNav isLoading={isLoading}>
```
richarddubay: Then I passed that prop to the secondary nav here. I did this part in multiple places.

src/group-finder/GroupSingle.js
```diff
@@ -95,162 +90,193 @@ GroupInfo.propTypes = {
...
+                    },
+                  ]}
+                  linkToGroup={false}
+                />
+              </Right>
+            </MediaQuery>
+          ) : null}
+        </FlexedSideBySideView>
+        <MediaQuery maxWidth="md">
+          <SecondaryNav isLoading={isLoading}>
```
richarddubay: Here.

src/music/Playlist.js
```diff
@@ -16,52 +16,48 @@ const enhance = compose(
...
+            isLoading={isLoading}
+            title={title}
+            artist="NewSpring"
+            albumImage={getAlbumImageSource(images)}
+            blurredImage={getBlurredImageSource(images)}
+          />
+        }
+      />
+      <MediaQuery maxWidth="md">
+        <SecondaryNav isLoading={isLoading}>
```
richarddubay: and here.

src/news/Single.js
```diff
@@ -20,35 +20,33 @@ const enhance = compose(
...
+      <Header titleText="News" backButton />
+      <ScrollView>
+        <ContentView isLoading={isLoading} {...otherContentProps}>
+          <Title>{startCase(toLower(title))}</Title>
+          <ByLine authors={authors} />
+          <HTMLView>{body}</HTMLView>
+        </ContentView>
+      </ScrollView>
+      <MediaQuery maxWidth="md">
+        <SecondaryNav isLoading={isLoading}>
```
richarddubay: and here

src/series/Sermon.js
richarddubay: and here

src/series/Single.js
```diff
@@ -48,56 +48,56 @@ const StyledButton = styled(({ theme }) => ({
...
         !isLoading && <RelatedContent tags={tags} excludedIds={[id]} />}
-    </ScrollView>
-    <SecondaryNav>
-      <ShareLink id={id} />
-      <Like id={id} isLiked={isLiked} />
-    </SecondaryNav>
-  </BackgroundView>
-));
+      </ScrollView>
+      <SecondaryNav isLoading={isLoading}>
```
richarddubay: and here

src/stories/Single.js
```diff
@@ -21,36 +21,35 @@ const enhance = compose(
...
+      <ScrollView>
+        <ContentView isLoading={isLoading} {...otherContentProps}>
+          <Title>{startCase(toLower(title))}</Title>
+          <ByLine authors={authors} />
+          <HTMLView>{body}</HTMLView>
+        </ContentView>
+        <RelatedContent tags={tags} excludedIds={[id]} />
+      </ScrollView>
+      <MediaQuery maxWidth="md">
+        <SecondaryNav isLoading={isLoading}>
```
richarddubay: and here

src/studies/Entry/index.js
```diff
@@ -41,58 +41,57 @@ const enhance = compose(
...
+              isLoading,
+            })(DevotionalTab),
+            scripture: withProps({
+              scripture,
+              entryData: children,
+              isLoading,
+            })(ScriptureTab),
+          })}
+        />
+        <SecondaryNav isLoading={isLoading}>
```
richarddubay: and here.

src/studies/Single.js
```diff
@@ -37,46 +37,40 @@ const enhance = compose(
...
         !isLoading && <RelatedContent tags={tags} excludedIds={[id]} />}
-    </ScrollView>
-    <SecondaryNav>
-      <ShareLink id={id} />
-      <Like id={id} isLiked={isLiked} />
-    </SecondaryNav>
-  </BackgroundView>
-));
+      </ScrollView>
+      <SecondaryNav isLoading={isLoading}>
```
richarddubay: and finally here.

richarddubay:
@conrad-vanl Yep! I added comments to pretty much all of the lines that I changed. Sorry about all the `prettier` stuff.

src/@ui/SecondaryNav/index.js
conrad-vanl: This is Ok, but be careful returning a conditional expression in a render function like this. You're better off doing something like:

```
!isLoading ? (<MyComponent ..../>) : null
```

because then you explicitly return `null` instead of `!isLoading`. The problem with returning a conditional expression like this is that you don't really want to render `!isLoading`, you want to render `null`. In this case, `!isLoading` would resolve to `false` and nothing would be rendered, which is why this is Ok. But say you omitted the `!` ... and did something like:

```
conditionalProp && <MyComponent />
```

If `conditionalProp` is ever an empty string, that's a falsey value, but still a string. So an empty string would be returned. React-Native would then try and render an empty string - which is invalid unless wrapped inside of a `<Text>` component. Therefore, you'd get a RSOD. You could do:

```
!!conditionalProp && <MyComponent />
```

to make sure you're always returning a component or a boolean, but nonetheless: if you don't want to render something based off of a prop, it's better to make sure you return `null` as that is handled more consistently within react. Using ternaries like this (`someProp ? <RenderThis /> : null`) is the standard we've adopted in this project to assert render-ability.

richarddubay: @conrad-vanl This is super helpful and makes a ton of sense! I'll make that change.

PR #563, opened 2018-04-30
Make homepage on web the giving dashboard

### LINKED ISSUE #542: [WEB] Homepage change to Giving Dashboard
## Acceptance Criteria

- [ ] On web the home page should be the giving dashboard, and the current home feed should go away

conrad-vanl:
Fixes #542 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/src/tabs/Layout/index.js
+++ b/src/tabs/Layout/index.js
@@ -1,5 +1,6 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
+import { Platform } from 'react-native';
 import {
   withContext,
   compose,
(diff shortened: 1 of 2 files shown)


src/tabs/Layout/index.js
```diff
@@ -34,7 +35,7 @@ class TabBarLayout extends PureComponent {
         </BackgroundView>
         {!this.props.hideTabBar && (
           <TabBar>
-            <Link to="/" icon="logo" label="Home" />
+            <Link to={Platform.OS === 'web' ? '/give' : '/'} icon="logo" label="Home" />
```
cdanwards: @conrad-vanl do you mind explaining this in combination with the first file change? It looks redundant to me, but it may not be.

conrad-vanl: It makes the home button show an active state on web, and for some reason the redirect still wasn’t happening when pressing this button - I think possibly but with React router or react router doesn’t evaluate redirects when you navigate to a path from within the app

PR #571, opened 2018-05-03
Normalize Address State field across forms

### LINKED ISSUE #440: Address state field inconsistent and defaulting to SC
## Acceptance Criteria

- [x] The address state field should save the same type in all forms!
- [x] The address state field should have the same UI (dropdown) in all forms.
- [ ] The address state field (dropdown) should only display values relative to your country.

## Context

If the user has an address (state) defined in their user profile and try to give with a new payment method the state dropdown doesn't auto-populate with the correct state (always defaults to SC).
Additionally, these two forms (`BillingAddressForm` and `ProfileAddressForm`) save different data types for state. `BillingAddressForm` saves a `stateId` (number sometimes in the form of a string) and `ProfileAddressForm` saves `values.state` (state name in the form of a string). Lastly, the values shown in the state dropdown do not currently represent the country that is selected.

#### Billing Address Form
simulator screen shot - iphone x - 2018-03-29 at 09 18 49
#### Profile Address Form
simulator screen shot - iphone x - 2018-03-29 at 09 18 07
#### Preferred UI for State selection
simulator screen shot - iphone x - 2018-03-29 at 11 50 06

cdanwards:
Fixes #440 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

I added an input picker for state in the Profile AddressForm. With this change, I created a new data component, withAddressState, within withUser that can be used in BillingAddressForm so that changes made to Profile Address can be used elsewhere and not have to have withCheckout in the ProfileAddressForm.

### DIFF
--- a/src/@data/withCheckout/checkoutQuery.js
+++ b/src/@data/withCheckout/checkoutQuery.js
@@ -17,9 +17,9 @@ export default gql`
       lastName
       email
       campus { name, id: entityId }
-      home { street1, street2, city, state, zip, country }
+      home { id, street1, street2, city, state, zip, country }
     }
-    savedPaymentMethods: savedPayments(cache: false){
+    savedPaymentMethods: savedPayments(cache: false) {
       id: entityId
       name
       payment {
--- a/src/@data/withUser/withUpdateHomeAddress.js
+++ b/src/@data/withUser/withUpdateHomeAddress.js
@@ -27,6 +27,18 @@ export default graphql(MUTATION, {
     updateHomeAddress: (params = {}) =>
       mutate({
         variables: pick(params, ['street1', 'street2', 'city', 'stateId', 'zip']),
+        optimisticResponse: {
+          __typename: 'Mutation',
+          updateHomeAddress: {
+            id: params.id,
+            __typename: 'Location',
+            street1: params.street1,
+            street2: params.street2,
+            city: params.city,
+            stateId: params.stateId,
+            zip: params.zip,
+          },
+        },
       }),
   }),
 });
--- a/src/@data/withUser/withUpdateHomeAddress.js
+++ b/src/@data/withUser/withUpdateHomeAddress.js
@@ -6,7 +6,6 @@ export const MUTATION = gql`
   mutation updateHomeAddress(
     $street1: String
     $street2: String
-    $countryId: String
     $city: String
     $stateId: String
     $zip: String
@@ -15,9 +14,8 @@ export const MUTATION = gql`
       input: {
         Street1: $street1
         Street2: $street2
-        CountryId: $countryId
         City: $city
-        StateId: $stateId
+        State: $stateId
         PostalCode: $zip
       }
     )
@@ -28,7 +26,7 @@ export default graphql(MUTATION, {
   props: ({ mutate }) => ({
     updateHomeAddress: (params = {}) =>
       mutate({
-        variables: pick(params, ['street1', 'street2', 'countryId', 'city', 'stateId', 'zip']),
+        variables: pick(params, ['street1', 'street2', 'city', 'stateId', 'zip']),
       }),
   }),
 });
--- a/src/@data/withUser/withUpdateHomeAddress.js
+++ b/src/@data/withUser/withUpdateHomeAddress.js
@@ -4,32 +4,31 @@ import pick from 'lodash/pick';
 
 export const MUTATION = gql`
   mutation updateHomeAddress(
-    $street1: String,
-    $street2: String,
-    $city: String,
-    $state: String,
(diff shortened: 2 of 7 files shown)


conrad-vanl:
While I would love to approve this, since I worked on it we should probably get someone else to review it. @burkeshartsis, @IsaacHardy, or @richarddubay have at it.

src/@data/withCheckout/checkoutQuery.js
```diff
@@ -17,9 +17,9 @@ export default gql`
       lastName
       email
       campus { name, id: entityId }
-      home { street1, street2, city, state, zip, country }
+      home { id, street1, street2, city, state, zip, country }
```
conrad-vanl: This `id` allows apollo to update `home` address optimistically

src/@data/withUser/withUpdateHomeAddress.js
```diff
@@ -4,32 +4,41 @@ import pick from 'lodash/pick';
...
+          __typename: 'Mutation',
+          updateHomeAddress: {
+            id: params.id,
+            __typename: 'Location',
+            street1: params.street1,
+            street2: params.street2,
+            city: params.city,
+            stateId: params.stateId,
+            zip: params.zip,
+          },
```
conrad-vanl: This optimistic response was added because Apollo isn't properly updating the store for a person without it after updating their home address

src/@ui/forms/BillingAddressForm.js
```diff
@@ -158,17 +159,18 @@ const validationSchema = Yup.object().shape({
...
-  city: get(props, 'contributions.city') || get(props, 'person.home.city', ''),
-  stateId: get(props, 'contributions.stateId') || get(props, 'person.home.state') || 'SC',
-  countryId: get(props, 'contributions.countryId') || get(props, 'person.home.country') || 'US',
-  zipCode: get(props, 'contributions.zipCode') || get(props, 'person.home.zip', ''),
+  street1: get(props, 'user.home.street1', ''),
+  street2: get(props, 'user.home.street2', ''),
+  city: get(props, 'user.home.city', ''),
+  stateId: get(props, 'user.home.state') || 'SC',
+  countryId: get(props, 'user.home.country') || 'US',
+  zipCode: get(props, 'user.home.zip', ''),
```
conrad-vanl: @cdanwards can you remind me why these fields were changed so much? I believe we still need the `get(props, 'contributions.XXXXX') || ` stuff - otherwise if a user fills out their billing address, goes to the next step, and then goes back - I believe this form will default back to their profile address, instead of their previously entered billing address.

Furthermore, I don't believe `withUser` is needed at all here.

cdanwards: @conrad-vanl I was typing out a response, then realized I had the requirements of the issue confused. So, for my clarification, do we want to have the billing address form of a new payment method prefilled with the values from the profile address? If not, then yes we can absolutely change these back and remove withUser.

IsaacHardy:
@cdanwards Let's try to solve the last issue in issue #440 in this PR. I will create an enhancement issue for pre-filling the new payment form with existing user address.

src/@ui/forms/BillingAddressForm.js
```diff
@@ -117,17 +119,30 @@ export const BillingAddressFormWithoutData = enhance(
...
                 value={values.stateId}
                 displayValue={get(states.find(state => state.id === values.stateId), 'label')}
                 onValueChange={value => setFieldValue('stateId', value)}
                 error={touched.stateId && errors.stateId}
               >
-                {states.map(({ label, id }) => (
-                  <Inputs.PickerItem label={label} value={id} key={id} />
-                ))}
+                {states
+                  .filter(state => state.idNumber >= 69 && state.idNumber <= 127)
```
IsaacHardy: Instead of inserting these ids so bluntly, lets create a variable above that is more descriptive of these id numbers.

With that being said, I'd prefer we update this on Rock's end so this change isn't needed at all.

richarddubay: Even if we update the defined type/value in Rock, we would still need to pull the data for both the states and provinces. We just wouldn't need the hardcoded id's. Which is fine. I'd rather not hard code things anyway. We'd have to update the query to pull back data for both ... which shouldn't be too difficult.

richarddubay:
So that it's written down...what was decided was that we would pull out the changes that we merged for splitting the states and territories because it hardcoded values. What @IsaacHardy would like to do is to hardcode the values and do the splitting in Heighliner and send back the data separately. Then we can make another change to get that data and correct the dropdown once that is changed.

PR #574, opened 2018-05-04
Fix logout promise fullfillment and logout redirection on web

### LINKED ISSUE #540: [WEB] Investigate logout
## Acceptance Criteria

- [ ] When logging out on web, we should redirect the user to the home screen and make sure state is cleared

conrad-vanl:
Fixes #540 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [x] Review new test logic

### DIFF
--- a/src/@data/withUser/withLogout.js
+++ b/src/@data/withUser/withLogout.js
@@ -16,7 +16,7 @@ export default graphql(MUTATION, {
       try {
         const r = await mutate();
         await AsyncStorage.removeItem('authToken');
-        await clearCache();
+        clearCache();
         return r;
       } catch (err) {
         throw err;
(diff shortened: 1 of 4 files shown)


src/@data/withUser/withLogout.js
```diff
@@ -16,7 +16,7 @@ export default graphql(MUTATION, {
       try {
         const r = await mutate();
         await AsyncStorage.removeItem('authToken');
-        await clearCache();
+        clearCache();
```
conrad-vanl: clearCache doesn't return a promise

src/settings/index.js
```diff
@@ -22,6 +23,7 @@ const LogoutTouchable = compose(
   withProps(({ logout }) => ({
     async onPress() {
       await logout();
+      if (Platform.OS === 'web') reload();
```
conrad-vanl: `reload` does a better clear of store, and fixes the weird redirection loop that's going on with web currently. I'd like to do the same on mobile, but it currently causes a different bug.

PR #577, opened 2018-05-07
add formik setter HOCs for pure rendering of form inputs

### LINKED ISSUE #541: [WEB] Investigate poor form performance
Current web forms kinda suck. Sometimes text entry is slow, autofill seems to not work right, and since none of our form inputs are wrapped in `<form>` fields, a user can't press their enter key to submit a form.

conrad-vanl:
Fixes #541 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [x] Review new test logic


src/@ui/forms/BillingAddressForm/__snapshots__/BillingAddressForm.tests.js.snap
```diff
@@ -64,7 +64,7 @@ exports[`The ContributionForm component hides state field for foreign countries
                 allowFontScaling={true}
                 editable={true}
                 onBlur={[Function]}
-                onChangeText={[Function]}
+                onChangeText={undefined}
```
conrad-vanl: FYI: these got changed to `undefined`s due to the use of `jest.fn()` above

src/@ui/forms/ProfileAddressForm.js
```diff
@@ -64,12 +65,12 @@ const enhance = compose(
 
 export const ProfileAddressFormWithoutData = enhance(
   ({
-    setFieldValue,
+    createFieldTouchedHandler,
```
burkeshartsis: this change isn't represented in the proptypes above

src/@ui/forms/ProfileAddressForm.js
```diff
@@ -64,12 +65,12 @@ const enhance = compose(
...
     handleSubmit,
     values,
     countries,
     states,
-    setFieldTouched,
+    createFieldValueHandler,
```
burkeshartsis: this change isn't represented in the proptypes above

src/@ui/forms/formikSetters.tests.js
```diff
@@ -0,0 +1,64 @@
...
+    expect(tree).toMatchSnapshot();
+  });
+  it('to create a handler that calls setFieldTouched', () => {
+    const setter = jest.fn();
+    const TestComponent = ({ createFieldTouchedHandler }) => {
+      expect(createFieldTouchedHandler).toBeDefined();
+      const handler = createFieldTouchedHandler('testField');
+      handler('someValue');
+      expect(setter.mock.calls.length).toBe(1);
+      return null;
```
burkeshartsis: The snapshot for this returns `null` but we assert `1` above. Should we return that value for clarity?

src/@ui/forms/formikSetters.tests.js
```diff
@@ -0,0 +1,64 @@
...
+    const transform = jest.fn().mockReturnValue('transform');
+    const TestComponent = ({ createFieldValueHandler }) => {
+      expect(createFieldValueHandler).toBeDefined();
+      const handler = createFieldValueHandler('testField', transform);
+      handler('someValue');
+      expect(setter.mock.calls.length).toBe(1);
+      expect(setter).toHaveBeenCalledWith('testField', 'transform');
+      expect(transform.mock.calls.length).toBe(1);
+      expect(transform).toHaveBeenCalledWith('someValue');
+      return null;
```
burkeshartsis: The snapshot for this returns `null` and doesn't match or map to our asserted outcome. What could we return that would be more meaningful?

src/@ui/forms/formikSetters.tests.js
burkeshartsis: The snapshot for this returns `null` but we assert `1` (`setter.mock.calls.length`) above. Should we return that value for clarity?

src/@ui/inputs/Picker/index.web.js
```diff
@@ -9,6 +9,8 @@ import InputWrapper from '../InputWrapper';
 import FloatingLabel from '../FloatingLabel';
 import withInputControlStyles from '../withInputControlStyles';
 
+const enhance = pure;
```
burkeshartsis: 😍

src/@ui/inputs/Picker/index.web.js
```diff
@@ -19,13 +21,13 @@ const StyledNativePicker = compose(
   }),
 )(NativePicker);
 
-const Picker = ({ wrapperStyle, value, ...props }) => (
+const Picker = enhance(({ wrapperStyle, value, ...props }) => (
   <InputWrapper style={wrapperStyle}>
     <StyledNativePicker {...props} selectedValue={value} prompt={props.placeholder} />
```
burkeshartsis: 🤔 Spreading all props blindly like this can cause unknown rerenders on a otherwise pure component because of props that are being passed down that you might not even be using. Might be worth refactoring this to be more explicit about the props needed or use `onlyUpdateForPropTypes` instead of `pure` (or both!)

src/@ui/forms/BillingAddressForm/GeographicPicker.js
```diff
@@ -0,0 +1,34 @@
...
+    value={value}
+    {...pickerProps}
+  >
+    {options.map(({ label, id }) => (
+      <Inputs.PickerItem label={label} value={id} key={id} />
+    ))}
+  </Inputs.Picker>
+));
+
+GeographicPicker.propTypes = {
```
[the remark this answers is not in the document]
conrad-vanl: I am no longer a fan of `setPropTypes` :( I'd love to suggest moving away from the practice of it all together. Main reason has to do with the fact that `propTypes` never becomes a property on the component, secondary has to do with `setPropTypes` actually works differently then React's `propType` validation.

src/@ui/forms/BillingAddressForm/index.js
```diff
@@ -65,77 +70,65 @@ const enhance = compose(
...
+    createFieldValueHandler,
+    createFieldTouchedHandler,
     handleSubmit,
-    values,
-    countries,
-    states,
-    setFieldTouched,
-    touched,
-    errors,
+    values = {},
```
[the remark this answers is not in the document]
conrad-vanl: we don't use `defaultProps` anywhere else in this component, nor in the other form components :/

src/@ui/inputs/Picker/index.web.js
conrad-vanl: this spread isn't new and I'd prefer to not change it in this PR (don't want to mess up something we don't have tests for yet)

src/@ui/forms/formikSetters.tests.js
conrad-vanl: took snapshots out, not needed

src/@ui/forms/formikSetters.tests.js
conrad-vanl: took snapshots out, not needed

src/@ui/forms/formikSetters.tests.js
conrad-vanl: took snapshots out, not needed

PR #614, opened 2018-05-16
Release v6.0.2 HOTFIX

IsaacHardy:
Removed sentry from catch handler for paymentform

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


IsaacHardy:
@FrankGrand @conrad-vanl @richarddubay This is a hotfix for our current giving issue. It seemed that Sentry was causing a memory leak whenever we had to log a new error from a user. This is a temporary fix while we look for a better solution moving forward.

PR #673, opened 2018-06-20
Add More Sentry

richarddubay:
Added more Sentry logging for more/better information on problems our people are seeing.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/src/@data/withGive/resolvers/mutations.js
+++ b/src/@data/withGive/resolvers/mutations.js
@@ -1,7 +1,7 @@
 import pick from 'lodash/pick';
 import { QUERY as contributionsQuery } from '@data/withGive/withContributions';
-import { INITIAL_STATE } from './queries';
 import sentry from '@utils/sentry';
+import { INITIAL_STATE } from './queries';
 
 export function addContribution(result, variables, { cache }) {
   const { contributions: state } = cache.readQuery({
--- a/src/@data/withGive/resolvers/mutations.js
+++ b/src/@data/withGive/resolvers/mutations.js
@@ -373,7 +373,7 @@ export function restoreContributions(result, variables, { cache }) {
   cache.writeQuery({
     query: contributionsQuery,
     data: {
-      contributions: JSON.parse(variables.state),
+      contributions: stateVariables,
     },
   });
   return null;
--- a/src/@data/withGive/resolvers/mutations.js
+++ b/src/@data/withGive/resolvers/mutations.js
@@ -363,7 +363,7 @@ export async function setSavedPaymentMethod(result, variables, { cache }) {
 }
 
 export function restoreContributions(result, variables, { cache }) {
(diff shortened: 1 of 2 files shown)


src/@utils/removeUndefined.js
```diff
@@ -1,5 +1,13 @@
...
 // module({ thing: undefined, data: 0 }) => { data: 0 }
+import sentry from '@utils/sentry';
+
 export default function (obj) {
-  return JSON.parse(JSON.stringify(obj));
+  let returnValue = {};
+  try {
+    returnValue = JSON.parse(JSON.stringify(obj));
+  } catch (e) {
+    sentry.captureException(e);
```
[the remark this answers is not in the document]
richarddubay: So the `removeUndefined` is used in `createValidationOrder` and `getOrderDetails`. I would agree that at least the `Invalid JSON` error we're already seeing _probably_ isn't being caused by this. I can remove that one if you think it's unnecessary.

src/@data/withGive/resolvers/mutations.js
```diff
@@ -374,12 +363,18 @@ export async function setSavedPaymentMethod(result, variables, { cache }) {
 }
 
 export function restoreContributions(result, variables, { cache }) {
+  let stateVariables = {};
+  try {
+    stateVariables = JSON.parse(variables.state);
+  } catch (e) {
+    sentry.captureException(e);
```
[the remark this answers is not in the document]
richarddubay: I believe this ends up being used in or near the `restoredCheckout` ... which is where the majority of the `Invalid JSON` errors seem to be coming from. So this feels different than the `removeUndefined` one. I'll see what I can do about possibly logging other information here and not _just_ the error.

richarddubay:
@conrad-vanl I was able to pull a little more data and actually bring back the `result` and the `variables`. It'll be interesting to see what we get with this.

src/@data/withGive/resolvers/mutations.js
```diff
@@ -374,12 +363,23 @@ export async function setSavedPaymentMethod(result, variables, { cache }) {
...
+        result,
+        variables,
+      },
+    });
+  }
+
   cache.writeQuery({
     query: contributionsQuery,
     data: {
-      contributions: JSON.parse(variables.state),
```
conrad-vanl: will this contain the user's creditcard / bankaccount info?

richarddubay: That's a great question ... that I don't know the answer to. I didn't look at what was in them, I just wanted to log them. I'll add some console logs and do a bit more looking and get back to you.

richarddubay: @conrad-vanl This should be all set now. Thanks for taking a look at it!

PR #686, opened 2018-08-15
Give date input

cdanwards:
Fixes [SYS-4451](https://newspring.atlassian.net/secure/RapidBoard.jspa?rapidView=58&projectKey=SYS&modal=detail&selectedIssue=SYS-4451)

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


android/app/src/main/java/host/exp/exponent/generated/DetachBuildConstants.java
```diff
@@ -3,6 +3,6 @@
 // This file is auto-generated. Please don't rename!
 public class DetachBuildConstants {
 
-  public static final String DEVELOPMENT_URL = "expba3ad9e96a6043fc95726c73cb381a27://t3-pit.newspring-builds.apollos.exp.direct:80";
+  public static final String DEVELOPMENT_URL = "expba3ad9e96a6043fc95726c73cb381a27://10.0.201.75:19000";
```
richarddubay: I kind of feel like this file shouldn't have been changed.

IsaacHardy: If this is reverted, make sure it still works locally.

cdanwards:
@richarddubay @IsaacHardy I've tested this PR again and it's GTG. Once y'all sign off on it we can OTA.

richarddubay:
@IsaacHardy Can we get you to review this? I made some changes so I don't feel great having me be the one to review it. Once you say yes, we'll start the deployment process. cc @cdanwards

IsaacHardy:
@richarddubay @cdanwards For the future, let's make sure we communicate the goal of the PR when we create them.

IsaacHardy:
We need to communicate more than just linking to an issue. It's important to communicate an understanding of what was change, how it was changed, and why it was changed. 

Also, it's also useful to demonstrate how to test or an example of the solution. That is already a listed item to created a PR.

IsaacHardy:
LGTM!

richarddubay:
@IsaacHardy I don't disagree, but I would suggest that this is not the way we currently write up PR's with the Apollos Project. Perhaps we should? If we want to do our PR's for our projects differently, that's fine.

IsaacHardy:
You are correct that we don't require that for a PR to be created, but many of us add these type of explanations when necessary. Its a case by case basis in The Apollos Project, but this type of practice is very important to communicate changes here at NewSpring. (Especially when you've been on vacation or at a tech conference and forgot everything regarding this PR)

richarddubay:
Haha! That's fair.

PR #692, opened 2018-09-14
AppCenter Implementation

IsaacHardy:
This connect's our Master builds to AppCenter so we can have CI for all merged pull requests.

AppCenter allows for:
* Automatic builds
* App Signing
* App Deployments
* App Testing
* App Distribution


android/app/src/main/java/host/exp/exponent/generated/DetachBuildConstants.java
```diff
@@ -3,6 +3,6 @@
 // This file is auto-generated. Please don't rename!
 public class DetachBuildConstants {
 
-  public static final String DEVELOPMENT_URL = "expba3ad9e96a6043fc95726c73cb381a27://t3-pit.newspring-builds.apollos.exp.direct:80";
+  public static final String DEVELOPMENT_URL = "expba3ad9e96a6043fc95726c73cb381a27://192.168.1.73:80";
```
richarddubay: Is this left over from your local build? I remember Dan had something like this change a couple of PR's back and he had to change it back.

ios/newspring.xcodeproj/xcshareddata/xcschemes/newspring.xcscheme
```diff
@@ -43,10 +56,9 @@
       </AdditionalOptions>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Debug"
+      buildConfiguration = "Release"
```
richarddubay: Should this be concerning?

index.ios.js
```diff
@@ -0,0 +1,4 @@
+import { AppRegistry } from 'react-native';
+import App from './App';
+
+AppRegistry.registerComponent('App', () => App);
```
richarddubay: Do we need to be this specific with `.ios` and `.android` files? They do the same thing.

android/app/src/main/assets/appcenter-config.json
```diff
@@ -0,0 +1,3 @@
+{
+    "app_secret": ""
+}
```
richarddubay: There is almost nothing here ... is this necessary?

richarddubay:
@IsaacHardy Also, I noticed that the tests failed because something to do with Coveralls. Shrug.

android/app/src/main/assets/appcenter-config.json
IsaacHardy: I do not know what this is... I'll investigate

index.ios.js
IsaacHardy: We absolutely need them

ios/newspring.xcodeproj/xcshareddata/xcschemes/newspring.xcscheme
IsaacHardy: Nope. When appcenter builds, we want it to use Release settings

android/app/src/main/java/host/exp/exponent/generated/DetachBuildConstants.java
IsaacHardy: It's quite possible. I'll investigate to make sure.

index.ios.js
IsaacHardy: The build scripts looks for the index file based on scope (e.g. web, ios, android). So we have to be a bit redundant to make sure AppCenter only pulls mobel builds.

PR #700, opened 2018-10-05
[SYS-4523] Ignore Expo Config Files

redreceipt:
Fixes [[SYS-4523](https://newspring.atlassian.net/browse/SYS-4523)]

The solution to avoid this minor annoyance is to use `git update-index --assume-unchanged <file>` for all the offending files. This will prevent them from being automatically staged when using `git commit -a`. 

I also removed the lines in the `.gitignore` file because for some reason they weren't being ignored anyway, hence the problem.

#700 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


ios/newspring/Supporting/EXBuildConstants.plist
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
IsaacHardy: Why do we need this file again?

.gitignore
```diff
@@ -23,12 +23,6 @@ Thumbs.db
 
 coverage/
 
-# generated at runtime
```
IsaacHardy: Why did we need to remove these from the `.gitignore` again?

redreceipt: This is the confusing part, they aren't being ignored anyway so I just assumed it'd be better to take them out of the file to avoid more confusion.

ios/newspring/Supporting/EXBuildConstants.plist
redreceipt: this is the file that Expo uses for the local iOS build to paste in the local dev url

redreceipt:
SYS-4523

redreceipt:
This one is dead. Let's just close it and bail. Not positive this is the best solution anyway, I understand way more about expo than I previously did so maybe I'll get motivated to look back into it later :)

@IsaacHardy give me a thumbs up and I'll close this PR without merging and delete the branch.

richarddubay:
@redreceipt I think this one is fine to close.

PR #705, opened 2018-10-22
Fix Android 6.1.0 Crashing

redreceipt:
Fixes [SYS-4545 Android Crashing from Firebase Call](https://newspring.atlassian.net/browse/SYS-4545)

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -21,7 +21,7 @@ android {
 
   defaultConfig {
     applicationId 'cc.newspring.newspringapp'
-    targetSdkVersion 25
+    targetSdkVersion 26
     versionCode 600030
     versionName '6.1.0'
     ndk {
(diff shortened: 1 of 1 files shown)


IsaacHardy:
@redreceipt Can you elaborate a bit on these changes? What are they doing? Why did this change need to be made? Etc..

redreceipt:
@IsaacHardy this fixes a Firebase lib conflict described here
https://stackoverflow.com/questions/46340333/unable-to-start-receiver-com-google-firebase-iid-firebaseinstanceidinternalrecei

From my vague understanding, the app is crashing because the older Firebase lib starts a service in the background which conflicts with Android 8.0.  It was set to Firebase-9.8.0 from ExpoKit. Updating to Firebase-10.2.1 has fixed issues for a lot of people but if that doesn't work, we could try the more complicated Android Intent fix in the described solution here:
https://stackoverflow.com/questions/51532059/android-gcm-wakefulbroadcastreceiver-startwakefulservice-crash-due-to-illegals

Lastly, if all that doesn't work we could roll back the build to SDK 25 (Android v7) but I don't remember why we had to upgrade it so that could cause adverse other adverse effects. Currently, we're on 26.

IsaacHardy:
@redreceipt Why are we currently using firebase?

redreceipt:
We’re not, react-native-maps is

IsaacHardy:
Are we using react-native-maps?

redreceipt:
Might not be exactly `react-native-maps` but it's some React Native library that uses an old Firebase lib. ExpoKit imports a bunch of libraries that we probably don't use in our code but they do.

android/app/build.gradle
```diff
@@ -21,7 +21,7 @@ android {
 
   defaultConfig {
     applicationId 'cc.newspring.newspringapp'
-    targetSdkVersion 26
+    targetSdkVersion 25
```
IsaacHardy: https://developer.android.com/distribute/best-practices/develop/target-sdk

Looks like this won't work...

PR #710, opened 2018-10-30
[SYS-4520] Fixes ActivityIndicator freeze after login on Give Now

cdanwards:
Fixes [SYS-4520](https://newspring.atlassian.net/secure/RapidBoard.jspa?rapidView=2&projectKey=SYS&modal=detail&selectedIssue=SYS-4520)

Issue:
When on a mobile browser, there were issues where when in the Give Now flow, when you logged-in when prompted, and you didn't have the auth cookies there for login, the `ActivityIndicator` would get stuck perpetually spinning.

Decisions Made:
After much frustration of trying to figure out the flow of the `isLoading` props through the different HOC's wrapped around the form and find where it's being mishandled or handed to one where it doesn't need it, I tried moving the `branch` where the `ActivityIndicator` was locating further into the HOC wrap. The trade off is that upon initial load after login, the `ActivityIndicator` spins for ~1 second longer, but the form underneath is actually able to load.

To Test:
Either in a browser on a simulator or on a browser mimicking a phone resolution, make sure to clear all cookies, and simply go to `my.newspring.cc/give` and enter an amount to give, then when prompted login as well.

Creator
-------

- [x] Set a relevant reviewer

Reviewer
--------

- [ ] Review new test logic


redreceipt:
Worked for me! But I don't understand enough to know if this is the best solution. @IsaacHardy ...

PR #711, opened 2018-10-31
Transaction Origination

IsaacHardy:
Client work to allow for the passing of transaction origination from either web, ios, or android.


IsaacHardy:
@richarddubay I don't want to merge until you have approved.

PR #713, opened 2018-11-07
(SYS-4562) Sync Alpha to Beta

richarddubay:
Fixes [issue-number]

Creator
-------

- [ ] Set a relevant reviewer

Reviewer
--------

- [ ] Review new test logic


redreceipt:
We need to figure out the state bug that doesn't pre-fill a giving account first

redreceipt:
Once #721 is merged into alpha, I think we're good here.

PR #716, opened 2018-11-09
[SYS-4551] Creates an alternate giving flow requiring login.

cdanwards:
Fixes [SYS-4551](https://newspring.atlassian.net/secure/RapidBoard.jspa?rapidView=2&projectKey=SYS&modal=detail&selectedIssue=SYS-4551)

Problem this fixes:
Currently, on the give now page, if you enter an amount to give before you sign in then that amount does not persist through login and you have to re-enter that amount. This will require sign in before an amount can be entered into the input, thuse working around the problem.

Decisions Made:
This is a new design for the flow of the give now screen. It doesn't functionally change much, because we made the decision in the initial development that we would not allow guest gifts, thus a login is required.

Creator
-------

- [x] Set a relevant reviewer

Reviewer
--------

- [ ] Review new test logic


IsaacHardy:
@cdanwards Conflicts in code need resolving.

cdanwards:
@IsaacHardy I fixed those merge conflicts. Should we decide to go with this flow, I believe it would negate the necessity of some of the other open PRs.

richarddubay:
@cdanwards Is this still needed? Does the problem still exist? There are some conflicts now that we've replaced branches that will need to be fixed if so.

IsaacHardy:
@cdanwards To add on to @richarddubay questions from the comment above, how do we stand on this PR? What do we need to do to move it forward?

PR #723, opened 2018-11-25
(SYS-4546) Fixes Giving Fund Name Being Too Large

redreceipt:
Removes the styling from the giving fund name (now matches Android) because it doesn't fit on iOS and Web currently and looks bad anyway. We could make a separate issue to make it look good if we want but for now, this is simply a bug fix.

screen shot 2018-11-25 at 3 10 02 pm

Creator
-------

- [ ] Set a relevant reviewer

Reviewer
--------

- [ ] Review new test logic


cdanwards:
@richarddubay I tested and this seems to be gtg, but just wanted your sign-off on this1

app.json
```diff
@@ -53,7 +53,7 @@
     },
     "extra": {
       "NODE_ENV": "development",
-      "COMMIT_SHA": "059f53b3e92863324c285a22a2444bce1860915a",
+      "COMMIT_SHA": "944cd769bc97a73cc62853997a5b4f8fb4176cc7",
```
richarddubay: Should this have gotten committed? I'm not sure anymore if this affects anything or not.

src/@ui/forms/ContributionForm/FundInput.js
```diff
@@ -29,24 +29,6 @@ const StyledTextInput = styled(
 const StyledPicker = styled(
   ({ theme }) => ({
     color: theme.colors.primary,
-    /* The styles below are pulled from H3 to match the surrounding text it's layed out with.
-   * TODO: refactor Picker to accept a text component
-   */
-    ...Platform.select({
-      ios: {
-        fontSize: theme.helpers.rem(1.8),
-        lineHeight: theme.helpers.verticalRhythm(1.8, 1.14),
```
richarddubay: I think these can just be changed to be the right styles. From my testing, 1.4 works for iOS.

src/@ui/forms/ContributionForm/FundInput.js
```diff
@@ -29,24 +29,6 @@ const StyledTextInput = styled(
...
-      },
-      /* RN android doesn't support setting these values at this time
-     * android: {
-     *   lineHeight: theme.helpers.verticalRhythm(1.8, 1.34),
-     * },
-     */
-      web: {
-        fontSize: theme.helpers.rem(1.8),
-        lineHeight: theme.helpers.verticalRhythm(1.8, 1.15),
```
richarddubay: Again, I think we should just change these to be the right styles. 1.6 works for the web.

redreceipt:
@richarddubay ok updated sizes

PR #730, opened 2018-12-06
[SYS-4571] Fixed end date to look at the end of the day

richarddubay:
Fixes SYS-4571

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/src/@data/withTransactions/index.js
+++ b/src/@data/withTransactions/index.js
@@ -8,7 +8,7 @@ const withTransactionsEngine = compose(
     setFilterDateRange: ({ setDateRange }) => ({ startDate, endDate }, format) => {
       setDateRange({
         startDate: startDate ? moment(startDate, format).format('MM/DD/YYYY') : '',
-        endDate: startDate ? moment(endDate, format).format('MM/DD/YYYY') : '',
+        endDate: startDate ? moment(endDate, format).format('MM/DD/YYYY 23:59:59') : '',
       });
     },
   }),


richarddubay:
To test this:
* On the `alpha` branch, run the app and go to giving history.
* See what the date is of the last transaction made.
* Click on the filter and choose a start date before the last transaction date and an end date that is the same date as the last transaction date.
* Cry a little inside when you see that the last transaction does not show up in the filtered list.
* Pull this new branch.
* Filter using the same dates
* Cheer that the filtered list of transactions now includes the last transaction! 
* Approve this pull request.

src/@data/withTransactions/index.js
```diff
@@ -8,7 +8,7 @@ const withTransactionsEngine = compose(
     setFilterDateRange: ({ setDateRange }) => ({ startDate, endDate }, format) => {
       setDateRange({
         startDate: startDate ? moment(startDate, format).format('MM/DD/YYYY') : '',
-        endDate: startDate ? moment(endDate, format).format('MM/DD/YYYY') : '',
+        endDate: startDate ? moment(endDate, format).format('MM/DD/YYYY 23:59:59') : '',
```
IsaacHardy: Interesting fix.

richarddubay: I hesitated with hardcoding the time. How do you feel about it?

IsaacHardy: I don't see a reason to hesitate. Its a logical way to use `moment` to pinpoint when you'd like the endDate to be.

IsaacHardy:
@richarddubay Having a hard time testing this. 1) The date picker won't allow me to pick dates in the past, and 2) I am having some very odd behavior take place when filtering.

richarddubay:
@IsaacHardy The date picker thing should be fixed with the above commit. Filtering to have a smaller list does bring back something like [SYS-4464](https://newspring.atlassian.net/browse/SYS-4464) into play. On super small lists the data just continually refetches ... which makes for an _interesting_ app experience.

IsaacHardy:
@richarddubay We need to make sure this entire experience doesn't have these interesting experiences within them. We should look into ways to fix the infinite refetch and the date picker/ filter

PR #748, opened 2018-12-20
Removes app.json From Repo

redreceipt:
Fixes [issue-number]

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic


IsaacHardy:
@redreceipt Could you further explain how this removes app.json and how our deployment process will change?

redreceipt:
> @redreceipt Could you further explain how this removes app.json and how our deployment process will change?

@IsaacHardy certainly. It doesn't remove `app.json`, that wasn't very clear. It removes it from the repo. The only reason we needed it before was for the `exp` calls in the development build process but I got around the need for that and the need for `exp` in the `prebuild.sh` file. It improves our process greatly in the three ways outlined in the description.

android/app/src/main/res/values/colors.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-  <color name="colorPrimary">#1b73b4</color>
-  <color name="colorPrimaryDark">#011A2D</color>
-  <color name="colorAccentLight">#1b73b4</color>
-  <color name="colorAccentDark">#FFFFFF</color>
+  <color name="colorPrimary">#6BAC43</color>
```
IsaacHardy: Were these colors added because of the removable of `app.json`?

redreceipt:
Those colors were added because I removed the iOS and Android sections of
app.json completely. Now where we get color styles are consistent and not
in two different locations.

--
Michael Neeley
Junior Web Developer
NewSpring Church
www.newspring.cc
Cell: (803) 622-6599

PR #749, opened 2019-01-09
(SYS-4582) Fix Scrolling Skips and Duplicates

redreceipt:
We are having problems across the app in many places with duplicates and skips and when scrolling. This attempts to rewrite the function that handles `fetchMore`. *Extensive* testing will need to be done.


IsaacHardy:
@richarddubay Since I helped with this, could you review it?

PR #757, opened 2019-02-04
(SYS-4605) Added NEW Short Link Logic

redreceipt:
*MUST* be tested on actual devices

Use this link to test on iOS and Android. Paste into a note or something and then click from there.

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review new test logic

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -92,7 +92,7 @@
 
         <data
             android:scheme="https"
-            android:host="rm2y5.app.goo.gl" />
+            android:host="newspring.page.link" />
       </intent-filter>
       <intent-filter>
         <action android:name="android.intent.action.VIEW" />
@@ -102,7 +102,7 @@
 
         <data
             android:scheme="http"
-            android:host="rm2y5.app.goo.gl" />
+            android:host="newspring.page.link" />
       </intent-filter>
     </activity>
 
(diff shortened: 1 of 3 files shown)


richarddubay:
> Use this link to test on iOS and Android. Paste into a note or something and then click from there.

What link?

How do I test this?

redreceipt:
@richarddubay https://newspring.page.link/M5Jz

richarddubay:
@redreceipt So I _tried_ testing this. But I think I failed miserably. I ran the app with the Expo link from `yarn start` and I ran the website from the IP address from `yarn web` ... both on my phone. I clicked the link you showed above from a note on my phone and it still opened up Safari first with the preview page. So either I didn't do something right ... or this didn't work as intended (the intention being that it would not have to open up Safari first and it would go right into the app as it does on Android).

What I'm wondering is if anything needs to happen in the `NewSpring/NewSpring` repo for the website in order to make this work. There is an `app-site-association` file there that handles the universal links for the app ... maybe something needs to happen there for the `newspring.page.link` URL to work as intended? Obviously, I have no clue if this is true ... just something I thought about.

Also obviously, there is a chance that I just tested this incorrectly. If that's true, let me know where I messed it up and I'd be happy to test it again. 👍

redreceipt:
I think you'd be better to load the app via xcode and studio on real
devices instead of trying to use a remote bundle. Since the code is in the
native files.

--
Michael Neeley
Junior Web Developer
NewSpring Church
www.newspring.cc
Cell: (803) 622-6599

richarddubay:
This seems to work just fine on iOS. There is currently an issue with Crashlytics on Android. See: https://stackoverflow.com/questions/55194687/no-signature-of-method-com-crashlytics-tools-gradle-crashlyticsplugin. Seems smart to just wait for Fabric to fix this that force us to hardcode the tools version.

PR #783, opened 2019-03-06
(SYS-4647) Adds Version Number to App

redreceipt:
This will allow us to see what version of the Expo bundle is loaded onto the app.


redreceipt:
Yea I’m not crazy about the solution but in the new app we won’t have to list the version anywhere because it’ll always be the same as the release :)

—
Michael

PR #785, opened 2019-03-24
(DEV-3623) Added Pre-selection Prop

redreceipt:
Fixes [issue-number]

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [x] Review new test logic

### DIFF
--- a/src/@ui/Meta/__tests__/__snapshots__/index.js.snap
+++ b/src/@ui/Meta/__tests__/__snapshots__/index.js.snap
@@ -75,7 +75,7 @@ Object {
       "name": "og:image",
     },
     Object {
-      "content": "https://my.newspring.cc",
+      "content": "http://localhost:3000",
       "name": "og:url",
     },
     Object {
@@ -191,7 +191,7 @@ Object {
       "name": "og:image",
     },
     Object {
-      "content": "https://my.newspring.cc",
+      "content": "http://localhost:3000",
       "name": "og:url",
     },
     Object {
@@ -285,7 +285,7 @@ Object {
       "name": "og:image",
     },
     Object {
-      "content": "https://my.newspring.cc",
+      "content": "http://localhost:3000",
       "name": "og:url",
     },
     Object {
@@ -379,7 +379,7 @@ Object {
       "name": "og:image",
     },
     Object {
-      "content": "https://my.newspring.cc",
+      "content": "http://localhost:3000",
       "name": "og:url",
     },
     Object {
--- a/src/@ui/Meta/__tests__/__snapshots__/index.js.snap
+++ b/src/@ui/Meta/__tests__/__snapshots__/index.js.snap
@@ -75,7 +75,7 @@ Object {
       "name": "og:image",
     },
     Object {
-      "content": "http://localhost:3000",
+      "content": "https://my.newspring.cc",
       "name": "og:url",
     },
     Object {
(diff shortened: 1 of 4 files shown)


IsaacHardy:
@redreceipt Looks like there's a linting issue.

src/@ui/Meta/__tests__/__snapshots__/index.js.snap
```diff
@@ -75,7 +75,7 @@ Object {
       "name": "og:image",
     },
     Object {
-      "content": "http://localhost:3000",
+      "content": "https://my.newspring.cc",
```
IsaacHardy: Let's keep this as localhost:3000

src/@ui/forms/ContributionForm/index.js
```diff
@@ -1,7 +1,14 @@
 import React, { Component } from 'react';
 import { View, Platform } from 'react-native';
 import PropTypes from 'prop-types';
-import { compose, withProps, branch, renderComponent, setPropTypes, defaultProps } from 'recompose';
+import {
```
IsaacHardy: Are there any changes in this file besides prettier updates? We should revert changes like these or not commit them in the future.

src/@ui/forms/ContributionForm/FundInput.js
```diff
@@ -73,6 +73,7 @@ const pickerWrapperStyle = {
 
 const FundInput = ({
   isFirst = false,
+  preselection = {},
   funds = [],
   value = {},
```
IsaacHardy: Is this used?

src/@ui/Meta/__tests__/__snapshots__/index.js.snap
redreceipt: Why? Seems like putting anything to do with local dev environments in the repo isn't right. Is that why the texts are failing?

src/@ui/forms/ContributionForm/FundInput.js
redreceipt: Yes it's used all throughout the file

src/@ui/forms/ContributionForm/index.js
redreceipt: Yes L#204 and L#210 are the biggest changes but most of it is prettier

src/@ui/Meta/__tests__/__snapshots__/index.js.snap
IsaacHardy: This is a snapshot. So it doesn't matter what is in these, but constantly changing them can get tedious.

IsaacHardy: You have a linting issue. That is why they are failing.

richarddubay: The hard thing about this snapshot is that it is pulling based off an environment variable, right? So whatever Travis has as the variable is what we should match against. Unless the variable in Travis needs to change.

redreceipt:
@IsaacHardy not sure why those snapshots are failing. they aren't failing locally.

IsaacHardy:
@redreceipt Your snapshots aren't failing. Your linting is failing.

src/@ui/Meta/__tests__/__snapshots__/index.js.snap
IsaacHardy: That's right. The is being pulled from the `.env` file.

IsaacHardy:
It looks good on iOS and web, but I'm having difficulties with my Android env.

IsaacHardy:
Looks good!

IsaacHardy:
I'd like @richarddubay or @cdanwards to review before we merge.

